PR #1, opened 2024-11-18
main: initial version of `image-builder` with basic `--list-images`

mvo5:
This commit adds the new `image-builder` binary (name still a bit TBD but this is the best so far IMHO). This binary is meant to build images from the CLI without the need to setup a daemon. The main use-case is CI/CD and admins running this in scripts or ad-hoc. The CLI should be pleasant to use.

This first commit adds the `list-images` command which is a thin wrapper around functionality from the `osbuild/images` library.

It will list all buildable images by default and can be trimmed down further via `--filter` which supports the filtering from the `images` library, see https://github.com/osbuild/images/pull/1015

It also supports `--output` which will output the result in the given format. Currently "text" and "json" are supported.

Note that this will not work on it's own yet, it will need an installed image-builder to get the repositories. This will need to get fixed via either:
1. a dependency package for `image-builder` that carries all the repos
2. a shared repo that contains the repos
3. using go:embed to get them (see images#1038)

[edit: rename from ibuilder->image-builder after the most recent discussion during the standup, while there was no consensus  it seems people did like image-builder a bit better]

Part of. COMPOSER-2395


cmd/image-builder/main.go
```diff
@@ -0,0 +1,66 @@
...
+	rootCmd.SetErr(osStderr)
+
+	listImagesCmd := &cobra.Command{
+		Use:          "list-images",
+		Short:        "List buildable images, use --filter to limit further",
+		RunE:         cmdListImages,
+		SilenceUsage: true,
+	}
+	listImagesCmd.Flags().StringArray("filter", nil, `Filter distributions by a specific criteria (e.g. "type:rhel*")`)
+	listImagesCmd.Flags().String("output", "", "Output in a specific format (text, json)")
```
ondrejbudai, line 55: I wonder if this should become an argument of `rootCmd` in the future?

cmd/image-builder/repos.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"github.com/osbuild/images/pkg/reporegistry"
+)
+
+// XXX: copied from "composer", should be exported there so
+// that we keep this in sync
+// XXX2: means we need to depend on osbuild-composer-common or a new rpm
+// that provides the relevant packages *or* we use go:embed (cf images#1038)
+var repositoryConfigs = []string{
```
ondrejbudai, line 11: Food for thought: Should we stop using our custom repo format, and start using the same format as `dnf` uses in `/etc/yum.repos.d`? Our depsolver should be already able to consume via the `root` opion. Definitely not something to address in this PR, but I think it's a good conceptual dicusssion.

go.mod
```diff
@@ -0,0 +1,121 @@
+module github.com/osbuild/image-builder-cli
+
+go 1.23
```
ondrejbudai, line 3: Should we match the version of osbuild/images? I like to use always the greatest and latest, but I would like to prevent any future issues with distros having too old versions.

ondrejbudai (commented):
Awesome! I put some "high-level" thoughts inline, but nothing is blocking (well, maybe the `ibuilder` one). I guess we should convert them to issues/discussions at some point?

go.mod
mvo5: I have no strong opinion here (and flip-flopped a bit when adding this), I'm happy to move to the version from images. I also hope that we will distribute this (at least optionally) similar to bib in a container so we have more freedom) - but for now let's be convervative and move to the images version

cmd/image-builder/main.go
mvo5: A really good questions, probably yes - most verbs will have an --output switch (list-images, build). Not entirely sure about manifest though.

cmd/image-builder/repos.go
mvo5: I have no strong opinion here, I would like to centralize what we have though. I would like to be able to build all supported imagetypes from this binary, ideally without much/any external dependencies. I was thinking go:embed for this and having the list of supported reos in images and the ability to override. But there is some pushback for this idea (not sure why though, I'm probably missing some details why this would be an issue).

thozza (approved):
This LGTM as the initial PR. :+1:

The first commit still contains a mention of `ibuilder`. I've added a few comments, but none of them should be considered blockers at this stage of development.

PR #2, opened 2024-11-18
list-images: add --datadir override for e.g. custom repositories

mvo5:
[draft as it is build on top of #1]

[I'm not 100% certain about this tbh, it feels a bit inelegant but it seems a super common request to be able to have a --repositores overrride switch - I initially added --repositories but that feels a bit akward as it needs to point to the *parent* of the repositories/ directory so I name it --datadir. I guess alternatively we could define a `--repository /path/to/repositories/actual-file.json` and leave it to the user to pick the right one, I'm not sure which is nicer as I'm not sure what the use-case is]

This commit adds a `--datadir` override that allows overriding
the default search paths for e.g. the custom repositories. In
practise only repository overrides are supported because that
is the only data stored there.


mvo5:
> Do we expect the datadir to contain other types of data in the future, beyond just repositories? If this is strictly meant for repo configs, maybe the CLI flag name should reflect that?

Thank you, it's a very good question. I started with --repository-dir but then "images"  in "reporegistry" will automatically append "repositories" to the "conf-path" (c.f. https://github.com/osbuild/images/blob/main/pkg/reporegistry/repository.go#L21) - which makes it a bit awkward because this naive usage  `--repository-directory ~/devel/osbuild-composer/repositories` would not work. It would have to be `--repository-directory ~/devel/osbuild-composer/` which is (to me) a bit unexpected. So I went with `datadir`, I guess the alternative would be to switch images so that it does not append `repositories` automatically and fix the consuemrs (but it would be an API break).

achilleas-k:
> It would have to be `--repository-directory ~/devel/osbuild-composer/` which is (to me) a bit unexpected. So I went with `datadir`, I guess the alternative would be to switch images so that it does not append `repositories` automatically and fix the consuemrs (but it would be an API break).

ah, I see.  In that case `--datadir` makes sense.  It also leaves the door open for other things, which I think is a good idea.

I don't have any good ideas for what else it might include.  Perhaps some defaults for how osbuild is run in a `osbuild.conf` file.  Not important right now though :)

achilleas-k (approved):

PR #4, opened 2024-11-27
image-builder-cli: work in progress draft with all current features

mvo5:
This is here to show what direction we take, not so that it actually gets merged. The workflow is that commits out of this branch will be cleaned up, opened for review and then this branch is rebased on the new "main" - but is is often useful to refer back to the "big picture" in these individual PRs and that can be found here.


mvo5:
This can be closed now, #8 contains all of it (except commit  [14b996](https://github.com/osbuild/image-builder-cli/pull/4/commits/414b996a22f8d364210cbcc082cf149c99fc7601) but I'm not sure we want this).

PR #6, opened 2024-12-02
cmd: implement manifest command

mvo5:
This commit implements the `manifest` command for `image-builder`.
It will generate an osbuild manifest based on the given inputs,
e.g.:
```console
$ ./image-builder manifest --blueprint /path/to/blueprint.toml centos-9 qcow2
{"version":"2","pipelines":[{"name":"build","runner":",...
```

Note that there is an integration test but because of the depsolve
it will be slow. It will be skipped when doing `go test -short`.

Integration tests for ostree commit, this is because  ostree commit integration tests will need "build" first because the way this is tested today is that first an edge-commit is created, then an edge-ami based on this commit is generated. But "build" is currently missing

Note that this has no README.md update for `manifest` because manifest is considered an internal command, once "build" is added (which is easy once manifest is there) "build" will be documented.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,42 @@
+# Hacking on image-builder-cli
+
+Hacking on `image-builder` should be fun and is easy.
+
+We have unit tests and some integration testing.
+
+## Setup
+
+To work on bootc-image-builder one needs a working Go environment. See
+[go.mod](go.mod).
+
+To run the testsuite install the test dependencies as outlined in the
+[github action](./.github/workflows/go.yml) under
+"Install test dependencies".
+
+## Code layout
+
+The go source code of image-builder-cli is under
+`./cmd/image-builder`. It uses the
+[images](https://github.com/osbuild/images) library internally to
+generate the images. Unit tests (and integration tests where it makes
+sense) are expected to be part of every PR but we are happy to help if
+those are missing from a PR.
+
+## Build
+
+Build by running:
+```console
+$ go build ./cmd/image-builder/
+```
+
+## Unit tests
+
+Run the unit tests via:
+```console
+$ go test -short ./...
+```
+
+There are some integration tests that can be run via:
+```console
+$ go test ./...
+```
--- a/cmd/image-builder/distro.go
+++ b/cmd/image-builder/distro.go
@@ -1,0 +1,30 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/osbuild/images/pkg/distro"
+)
+
+var distroGetHostDistroName = distro.GetHostDistroName
+
+// findDistro will ensure that the given distro argument do not
+// diverge. If no distro is set via the blueprint or the argument
+// the host is used to derive the distro.
+func findDistro(argDistroName, bpDistroName string) (string, error) {
+	switch {
+	case argDistroName != "" && bpDistroName != "" && argDistroName != bpDistroName:
+		return "", fmt.Errorf("error selecting distro name, cmdline argument %q is different from blueprint %q", argDistroName, bpDistroName)
+	case argDistroName != "":
+		return argDistroName, nil
+	case bpDistroName != "":
+		return bpDistroName, nil
+	}
+	// nothing selected by the user, derive from host
+	distroStr, err := distroGetHostDistroName()
+	if err != nil {
+		return "", fmt.Errorf("error deriving host distro %w", err)
+	}
+	fmt.Fprintf(osStderr, "No distro name specified, selecting %q based on host, use --distro to override", distroStr)
+	return distroStr, nil
+}
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -41,19 +41,32 @@ func cmdManifest(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	blueprintPath, err := cmd.Flags().GetString("blueprint")
+	archStr, err := cmd.Flags().GetString("arch")
+	if err != nil {
+		return err
+	}
+	if archStr == "" {
+		archStr = arch.Current().String()
+	}
+	distroStr, err := cmd.Flags().GetString("distro")
 	if err != nil {
 		return err
 	}
 
-	distroStr := args[0]
-	imgTypeStr := args[1]
-	var archStr string
-	if len(args) > 2 {
-		archStr = args[2]
-	} else {
-		archStr = arch.Current().String()
+	var blueprintPath string
+	imgTypeStr := args[0]
+	if len(args) > 1 {
+		blueprintPath = args[1]
 	}
+	bp, err := blueprintload.Load(blueprintPath)
+	if err != nil {
+		return err
+	}
+	distroStr, err = findDistro(distroStr, bp.Distro)
+	if err != nil {
+		return err
+	}
+
 	res, err := getOneImage(dataDir, distroStr, imgTypeStr, archStr)
 	if err != nil {
 		return err
@@ -66,10 +79,6 @@ func cmdManifest(cmd *cobra.Command, args []string) error {
 	mg, err := manifestgen.New(repos, &manifestgen.Options{
 		Output: osStdout,
 	})
-	if err != nil {
-		return err
-	}
-	bp, err := blueprintload.Load(blueprintPath)
 	if err != nil {
 		return err
 	}
@@ -108,15 +117,15 @@ operating sytsems like centos and RHEL with easy customizations support.`,
 	rootCmd.AddCommand(listImagesCmd)
 
 	manifestCmd := &cobra.Command{
-		Use:          "manifest <distro> <image-type> [<arch>]",
+		Use:          "manifest <image-type> [blueprint]",
 		Short:        "Build manifest for the given distro/image-type, e.g. centos-9 qcow2",
 		RunE:         cmdManifest,
 		SilenceUsage: true,
-		Args:         cobra.MinimumNArgs(2),
+		Args:         cobra.RangeArgs(1, 2),
 		Hidden:       true,
 	}
-	// XXX: share with build
-	manifestCmd.Flags().String("blueprint", "", `pass a blueprint file`)
+	manifestCmd.Flags().String("arch", "", `build manifest for a different architecture`)
+	manifestCmd.Flags().String("distro", "", `build manifest for a different distroname (e.g. centos-9)`)
 	rootCmd.AddCommand(manifestCmd)
 
 	return rootCmd.Execute()
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -9,8 +9,8 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/osbuild/images/pkg/arch"
-	"github.com/osbuild/images/pkg/blueprint"
 
+	"github.com/osbuild/image-builder-cli/internal/blueprintload"
 	"github.com/osbuild/image-builder-cli/internal/manifestgen"
 )
 
@@ -41,6 +41,10 @@ func cmdManifest(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	blueprintPath, err := cmd.Flags().GetString("blueprint")
+	if err != nil {
+		return err
+	}
 
 	distroStr := args[0]
 	imgTypeStr := args[1]
@@ -65,8 +69,12 @@ func cmdManifest(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	var bp blueprint.Blueprint
-	return mg.Generate(&bp, res.Distro, res.ImgType, res.Arch, nil)
+	bp, err := blueprintload.Load(blueprintPath)
+	if err != nil {
+		return err
+	}
+
+	return mg.Generate(bp, res.Distro, res.ImgType, res.Arch, nil)
 }
 
 func run() error {
@@ -107,7 +115,8 @@ operating sytsems like centos and RHEL with easy customizations support.`,
 		Args:         cobra.MinimumNArgs(2),
 		Hidden:       true,
 	}
-	// XXX: add blueprint switch
+	// XXX: share with build
+	manifestCmd.Flags().String("blueprint", "", `pass a blueprint file`)
 	rootCmd.AddCommand(manifestCmd)
 
 	return rootCmd.Execute()
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -7,17 +7,17 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+
+	"github.com/osbuild/images/pkg/arch"
+	"github.com/osbuild/images/pkg/blueprint"
+
+	"github.com/osbuild/image-builder-cli/internal/manifestgen"
 )
 
 var (
 	osStdout io.Writer = os.Stdout
 	osStderr io.Writer = os.Stderr
 )
-
-type cmdlineOpts struct {
-	dataDir string
-	out     io.Writer
-}
 
 func cmdListImages(cmd *cobra.Command, args []string) error {
 	filter, err := cmd.Flags().GetStringArray("filter")
@@ -33,11 +33,40 @@ func cmdListImages(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	opts := &cmdlineOpts{
-		out:     osStdout,
-		dataDir: dataDir,
+	return listImages(dataDir, output, filter)
+}
+
+func cmdManifest(cmd *cobra.Command, args []string) error {
+	dataDir, err := cmd.Flags().GetString("datadir")
+	if err != nil {
+		return err
 	}
-	return listImages(output, filter, opts)
+
+	distroStr := args[0]
+	imgTypeStr := args[1]
+	var archStr string
+	if len(args) > 2 {
+		archStr = args[2]
+	} else {
+		archStr = arch.Current().String()
+	}
+	res, err := getOneImage(dataDir, distroStr, imgTypeStr, archStr)
+	if err != nil {
+		return err
+	}
+	repos, err := newRepoRegistry(dataDir)
+	if err != nil {
+		return err
+	}
+	// XXX: add --rpmmd/cachedir option like bib
+	mg, err := manifestgen.New(repos, &manifestgen.Options{
+		Output: osStdout,
+	})
+	if err != nil {
+		return err
+	}
+	var bp blueprint.Blueprint
+	return mg.Generate(&bp, res.Distro, res.ImgType, res.Arch, nil)
 }
 
 func run() error {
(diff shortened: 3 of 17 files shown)


internal/manifestgen/manifestgen.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	preManifest, warnings, err := imgType.Manifest(bp, *imgOpts, repos, 0)
+	if err != nil {
+		return err
+	}
+	if len(warnings) > 0 {
+		// XXX: what can we do here? for things like json output?
+		// what are these warnings?
+		return fmt.Errorf("warnings during manifest creation: %v", strings.Join(warnings, "\n"))
+	}
```
achilleas-k, line 156: We don't have many warnings but when we do they come from the `checkOptions()` call in each distro, which validates customizations for a given distro and image type.
The one warning I can point to (which I think is when warnings were introduced) is when a user is added in a blueprint for an ostree commit.  We wanted to deprecate that feature for RHEL 10, but now that edge is going away completely and bootc is a whole different thing, we should probably drop that.

Warnings are still probably useful for these kinds of things in the future though.

cmd/image-builder/main.go
```diff
@@ -69,6 +104,17 @@ operating sytsems like centos and RHEL with easy customizations support.`,
 	listImagesCmd.Flags().String("output", "", "Output in a specific format (text, json)")
 	rootCmd.AddCommand(listImagesCmd)
 
+	manifestCmd := &cobra.Command{
+		Use:          "manifest <distro> <image-type> [<arch>]",
```
achilleas-k, line 117: Would it be nicer if the architecture was a flag option (`--arch`)?
That way we could make blueprint an optional positional argument, which I think would be a more common usage.  Switching the arch would probably be less common.

cmd/image-builder/main.go
```diff
@@ -33,11 +33,46 @@ func cmdListImages(cmd *cobra.Command, args []string) error {
...
+	// XXX: this is not symetric, ideally we would only pass
+	// either a single reporegistry from a datadir everywhere
+	// or we would pass datadir everywhere but with just
+	// datadir we cannot create test distros as they cannot be
+	// defined outside of code so manifestgen needs to take a
+	// reporegistry instead of just a datadir
+	res, err := getOneImage(dataDir, distroStr, imgTypeStr, archStr)
+	if err != nil {
+		return err
+	}
```
achilleas-k, line 67: I'm not sure I completely understand the comment here.
Why can't test distros be defined outside?

HACKING.md
```diff
@@ -0,0 +1,41 @@
+# Hacking on image-builder-cli
+
+Hacking on `image-builder` should be fun and is easy.
+
+We have a unit tests some integration testing.
+
+## Setup
+
+To work on bootc-image-builder one needs a working Go [environment]. See
```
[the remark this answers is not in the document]
mvo5: Copy/pasted from bootc-image-builder (c.f. https://github.com/osbuild/bootc-image-builder/blob/main/HACKING.md#setup) sorry, silly me.

HACKING.md
```diff
@@ -0,0 +1,41 @@
...
+To run the testsuite install the test dependencies as outlined in the
+[github action](./.github/workflows/go.yml) under
+"Install test dependencies".
+
+## Code layout
+
+The go source code of bib is under `./cmd/image-builder`. It uses the
+[images](https://github.com/osbuild/images) library internally to
+generate the imagess. Unit tests (and integration tests where it
+makes sense) are expected to be part of a PR but we are happy to
```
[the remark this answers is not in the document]
mvo5: Thanks! I picked the second but no strong opinion

internal/manifestgen/manifestgen.go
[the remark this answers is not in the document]
mvo5: Thank you! Looks like I missed the final "!" the `fixup!` so when squashing it it not get auto-commented. Thanks for noticing. Fixed now.

internal/manifestgen/manifestgen.go
```diff
@@ -0,0 +1,176 @@
...
+type (
+	DepsolveFunc func(cacheDir string, packageSets map[string][]rpmmd.PackageSet, d distro.Distro, arch string) (map[string][]rpmmd.PackageSpec, map[string][]rpmmd.RepoConfig, error)
+
+	ContainerResolverFunc func(containerSources map[string][]container.SourceSpec, archName string) (map[string][]container.Spec, error)
+
+	CommitResolverFunc func(commitSources map[string][]ostree.SourceSpec) (map[string][]ostree.CommitSpec, error)
+)
+
+// Options contains the optional optons for the manifest generation.
+// For unset options defaults will be used.
```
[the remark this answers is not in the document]
mvo5: Heh, indeed, it's a bit awkward, fixed^Wimproved^Wchanged

internal/manifestgen/manifestgen.go
mvo5: Thanks, that is super useful! My worry about warnings like this is that if this runs in a CI/CD or a script many/most people will not see the warnings. I wonder if should error by default on warnings and add a `--warnings-okay` switch that would just put them on stderr (or of course the other way around, show them by default only and add `--strict/--Werror`. 

If we want to just show them right now I would tweak the code to:
```go
diff --git a/internal/manifestgen/manifestgen.go b/internal/manifestgen/manifestgen.go
index 4861658..a8b35ae 100644
--- a/internal/manifestgen/manifestgen.go
+++ b/internal/manifestgen/manifestgen.go
@@ -86,6 +86,7 @@ type (
 type Options struct {
        Cachedir          string
        Output            io.Writer
+       WarningsOutput    io.Writer
        Depsolver         DepsolveFunc
        ContainerResolver ContainerResolverFunc
        CommitResolver    CommitResolverFunc
@@ -96,6 +97,7 @@ type Options struct {
 type Generator struct {
        cacheDir string
        out      io.Writer
+       warn     io.Writer
 
        depsolver         DepsolveFunc
        containerResolver ContainerResolverFunc
@@ -114,6 +116,7 @@ func New(reporegistry *reporegistry.RepoRegistry, opts *Options) (*Generator, er
 
                cacheDir:          opts.Cachedir,
                out:               opts.Output,
+               warn:              opts.WarningsOutput,
                depsolver:         opts.Depsolver,
                containerResolver: opts.ContainerResolver,
                commitResolver:    opts.CommitResolver,
@@ -121,6 +124,9 @@ func New(reporegistry *reporegistry.RepoRegistry, opts *Options) (*Generator, er
        if mg.out == nil {
                mg.out = os.Stdout
        }
+       if mg.warn == nil {
+               mg.warn = os.Stderr
+       }
        if mg.depsolver == nil {
                mg.depsolver = defaultDepsolver
        }
@@ -150,9 +156,8 @@ func (mg *Generator) Generate(bp *blueprint.Blueprint, dist distro.Distro, imgTy
                return err
        }
        if len(warnings) > 0 {
-               // XXX: what can we do here? for things like json output?
-               // what are these warnings?
-               return fmt.Errorf("warnings during manifest creation: %v", strings.Join(warnings, "\n"))
+               // TODO: add --strict/--Werror option that errors here
+               fmt.Fprintf(mg.warn, "Warnings during manifest creation:\n%s", strings.Join(warnings, "\n"))
        }
        packageSpecs, _, err := mg.depsolver(mg.cacheDir, preManifest.GetPackageSetChains(), dist, a.Name())
        if err != nil {

```

My gut feeling for this is:
a) keep it strict and add --warnings-okay (name tbd)
b) do it in a followup to keep review focused

*But* I don't know how often those warnings happen in practice so maybe my gut-feeling is off and I'm happy to unblock this with the above diff too and add the "-Werror" later.

cmd/image-builder/main.go
mvo5: Thanks, that is an interesting idea. The reason for current layout is that it's designed to be copy/paste friendly from `list-iamges`. I.e. 
```console
$ image-builder list-images
...
rhel-10.0 type:oci arch:x86_64
$ image-builder manifest rhel-10.0 type:oci arch:x86_64
...
```
so we could tweak the list output to
```console
$ image-builder list-images
...
rhel-10.0 type:vhd --arch x86_64
rhel-10.0 type:vmdk --arch x86_64
rhel-10.0 type:oci --arch x86_64
$ image-builder manifest rhel-10.0 type:oci --arch x86_64
...
```
(which looks a bit uneven but not too bad and for scripts we provide --json output).

Or we could drop this property (copy/pastable). Or we make the third argument fuzzy, i.e. if it starts with "arch:" we consider it an architecture and otherwise its a blueprint. (or something else I forgot?)

WDYT?

cmd/image-builder/main.go
mvo5: Below is long and probably too much of-internal monologue explanation but you don't need to read it as I moved on from "test-distro" now and just use the real ones for testing. Different tradeoffs but easier to work with and arguably more realistic :)

---

Sorry, maybe I should remove this paragraph. The background is that I am trying to use the `test-distro-1` in my tests. And there are some steps for that with that and the distrofactory that I would love to simplify. 

Right now one needs to do:
1. distrofactory needs repositories to create distros, repositories are defined in json files
2. disrofactory also needs registered factory functions
3.  once distrofactory.Factory is created no new factories can be created
4. this means that for testing one needs to create a "distrofactory.NewTestDefault" (becuase (3))
5. it also means for testing one needs to put a custom datadir into the reporegistry with `test-distro-1.json` and empty fake architecture 

So tests that use test-distro-1 need to create a custom datadir and mock the distrofactory.NewDefault because one source of data is "datadir" and the other is registered factory functions in code. It would be nice to be able to just define a datadir and get my test distros but that would means we would have to have an (optional?) way to create distros from yaml (i.e. I would like to get rid of the need to mock distrofactory.NewDefault).

I guess the alternative is to just not use test-distro here and just use a "real" one which is easier to do as the default factory will work. I think I will try this.

mvo5: Fwiw, the paragraph is removed because test-distro is no loner used.

internal/manifestgen/manifestgen.go
achilleas-k: There's also the FIPS warning: https://github.com/osbuild/images/blob/7600a5ae1c1e179e06d1cca3a97c7a464a6b9be7/pkg/distro/fedora/imagetype.go#L447-L450

cmd/image-builder/main.go
achilleas-k: > While that makes typing out a manifest or build command by hand a tiny bit more tedious, it also makes it nice that the args aren't positional anymore and opens up the way to make distro optional too.

One issue with this is that we're inventing a new convention for non-positional arguments with `option:value` instead of the standard `--option=value` ...

cmd/image-builder/main.go
achilleas-k: Thanks for the explanation :)

internal/manifestgen/manifestgen.go
mvo5: > I think a `-Werror` style option would be better but it depends what kind of use cases we're targeting and thinking about. You make a good point about warnings not being visible when the build is part of an automated pipeline, but I also think it'd be annoying for builds to break on a deprecation warning.
[..]

I guess it's hard to say which is better, from my PoV there are two cases:
1. the build was using something deprecated so from day-0 and the user is aware as the first run(s) where looked at, so if the deprecated thing stops working after say 2y that is no surprise
2. the build is using something that was fine on day-0 but we deprecated 1y in, no-one (probably) looks at the build logs because it's churning along fine for 1y. In 2y we remove the feature and the build fails, now people need to panic because their feature no longer works and is no longer available

Alternatively if we error by default it would be:
1. on day-0 the user puts in `--allow-warnings` no surprise here
2. he build is using something that was fine on day-0 but we deprecated 1y in, now it break. this is annoying and the user needs to change the job to use `--allow-warnings` (which we hopefully made clear from the error message). But they have 1y to plan how to replace the feature before it really goes away.

Of course this is all conjecture on my part, idk if people never look at their CI logs and I'm happy to just show warnings by default and add `-Werror` later (i.e. just apply the patch) [just wanted to eloborate on my gut feeling a little bit]

cmd/image-builder/main.go
mvo5: [..]
> > Or we could drop this property (copy/pastable).
> 
> It's a nice property but yes I don't think it's important enough to make the output of another command so unusual.
[..]

Let's do that then, it's the most simple and straightforward change. I am a little bit sad about it because now:
```console
$ image-builder manifest distro:rhel-10.0 type:oci arch:x86_64
error: cannot find blueprint "arch:x86_64"

$ image-builder build distro:rhel-10.0 type:oci arch:x86_64
error: cannot find blueprint "arch:x86_64"
```

We could change the default for `image-builder-cli list-image` to only show same arch images by default and hint at something like "--all" to see all, then we would have preserved the copy/paste property again (at least by default) and it's nicer for the "build" case. Open for ideas of course but that seems a reasonable path forward(?).

internal/manifestgen/manifestgen.go
achilleas-k: You make a good argument there.  I think we need to think about it some more, but there's definitely some value for the cases you described.

We could even go as far as having a separate option for each warning (`--ignore-warning=fips-host`), so that users need to make explicit decisions about which things they need to worry about changing.

internal/blueprintload/blueprintload.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/BurntSushi/toml"
+
+	"github.com/osbuild/images/pkg/blueprint"
```
thozza, line 12: I guess that the ship has already sailed in BIB world, but the initial intention was for `images/pkg/blueprint` not to be used to decode user-provided blueprints (as it's the case in osbuild-composer).

thozza (commented):
Nice work. I've added a few comments.

internal/blueprintload/blueprintload.go
mvo5: Yeah, that ship has sailed unfortunately /o\. Sorry, I just wasn't aware of this.

internal/manifestgen/manifestgen.go
```diff
@@ -0,0 +1,178 @@
...
+func (mg *Generator) Generate(bp *blueprint.Blueprint, dist distro.Distro, imgType distro.ImageType, a distro.Arch, imgOpts *distro.ImageOptions) error {
+	if imgOpts == nil {
+		imgOpts = &distro.ImageOptions{}
+	}
+
+	repos, err := mg.reporegistry.ReposByImageTypeName(dist.Name(), a.Name(), imgType.Name())
+	if err != nil {
+		return err
+	}
+	preManifest, warnings, err := imgType.Manifest(bp, *imgOpts, repos, 0)
```
ondrejbudai, line 150: What about the seed here? I think the we should make it random by default. A fixed seed should be opt-in.

ondrejbudai (commented):
Looks great. I would like to have an answer for the seeding, other comments are nitpicks. :)

internal/manifestgen/manifestgen.go
```diff
@@ -0,0 +1,178 @@
...
+// into a common helper in "images" or images should do this on its
+// own
+func defaultDepsolver(cacheDir string, packageSets map[string][]rpmmd.PackageSet, d distro.Distro, arch string) (map[string][]rpmmd.PackageSpec, map[string][]rpmmd.RepoConfig, error) {
+	solver := dnfjson.NewSolver(d.ModulePlatformID(), d.Releasever(), arch, d.Name(), cacheDir)
+	depsolvedSets := make(map[string][]rpmmd.PackageSpec)
+	repoSets := make(map[string][]rpmmd.RepoConfig)
+	for name, pkgSet := range packageSets {
+		res, err := solver.Depsolve(pkgSet, sbom.StandardTypeNone)
+		if err != nil {
+			return nil, nil, err
```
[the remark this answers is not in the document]
mvo5: Good point, I wrapped all errors from the default*Resolver now

internal/manifestgen/manifestgen.go
mvo5: Thanks, (very) nice catch and *urg* for missing this. 

Thinking a bit about it I wonder^Wthink we should look into images and tweak the API here, it seems in almost all "real" use-cases one would want a non-zero seed here so maybe this could become either a pointer (and on nil we just pick a random seed) or keep as is but assign a random value if we get "0" in images. Obviously followup material and I fixed this instance of the issue here but thinking about the future/a followup, wdyt?

cmd/image-builder/main.go
mvo5: Fwiw, I have no strong opinion, I like the copy/paste property of the original design and "bib" also rquires an argument (currently) to pass a blueprint. Going back to the previous way is as easy as removing the last commit :) But I'm more interested in a way forward right now than which way it is :) (I would love to get this in before EOY)

mvo5: This had some more (offline) discussion with Ondrej and Achilleas and the latest agreement is:
```
This commit tweaks the argument handling based on the suggestion
by Achilleas and Ondrej (thanks!). Now it takes

$ image-builder manifest qcow2 ./path/to/blueprint
...
$ image-builder manifest --arch s390x --distro centos-9 qcow2
...

If no arch is specified the hostname arch is used. If no distro
is specified in either the blueprint or the commandline it is
auto-detected based on the host.

Note that if the distro from the comandline and the blueprint
diverge an error is raised. We can relax this rule and just
add precedence, e.g. commandline always overrides the blueprint
but ideally we would have a clear use-case here so we start
conservative and can always relax this rule later (the inverse
is much harder).

This means it is no longer copy/paste friendly from `list-images`
by default but instead we can provide a new
`list-images --output=shell` option that outputs in exactly the
format that `image-builder [manifest|build]` need.
```

cmd/image-builder/distro.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+
+	"github.com/osbuild/images/pkg/distro"
+)
+
+var distroGetHostDistroName = distro.GetHostDistroName
+
+// findDistro will ensure that the given distro argument do not
+// diverge. If no distro is set via the blueprint or the argument
+// the host is used to derrive the distro.
```
ondrejbudai, line 13: I suggest running `s/derriv/deriv` on the whole file . :)

ondrejbudai (changes requested):
I like this very much. Some minor issues inline.

cmd/image-builder/main.go
```diff
@@ -9,14 +9,15 @@ import (
...
-	"github.com/osbuild/images/pkg/blueprint"
 
+	"github.com/osbuild/image-builder-cli/internal/blueprintload"
 	"github.com/osbuild/image-builder-cli/internal/manifestgen"
 )
 
 var (
-	osStdout io.Writer = os.Stdout
-	osStderr io.Writer = os.Stderr
+	osStdin  io.ReadCloser = os.Stdin
```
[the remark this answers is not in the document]
mvo5: Nice catch, not anymore since things moved into blueprintloader

cmd/image-builder/main.go
```diff
@@ -109,15 +118,15 @@ operating sytsems like centos and RHEL with easy customizations support.`,
 	rootCmd.AddCommand(listImagesCmd)
 
 	manifestCmd := &cobra.Command{
-		Use:          "manifest <distro> <image-type> [<arch>]",
+		Use:          "manifest <image-type> [blueprint]",
 		Short:        "Build manifest for the given distro/image-type, e.g. centos-9 qcow2",
 		RunE:         cmdManifest,
 		SilenceUsage: true,
-		Args:         cobra.MinimumNArgs(2),
+		Args:         cobra.MinimumNArgs(1),
```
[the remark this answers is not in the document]
mvo5: Nice, TIL!

ondrejbudai (approved):
Looks good from my PoV

supakeen (approved):

PR #7, opened 2024-12-03
main: list-images takes no args (just --filter)

mvo5:
Tiny commit to error if extra arguments are passed to list-images. Only `--filter` is supported there currently.


ondrejbudai (approved):
Thanks!

PR #8, opened 2024-12-03
cmd: add new `build` command

mvo5:
This commit adds the `build` command. It takes the same flags as
`manifest` and will build the specified image.

E.g.
```console
$ image-builder-cli build --distro centos-9 qcow2
...
```

### DIFF
--- a/internal/testutil/testutil.go
+++ b/internal/testutil/testutil.go
@@ -1,0 +1,62 @@
+package testutil
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type MockCmd struct {
+	binDir string
+	name   string
+}
+
+func MockCommand(t *testing.T, name string, script string) *MockCmd {
+	mockCmd := &MockCmd{
+		binDir: t.TempDir(),
+		name:   name,
+	}
+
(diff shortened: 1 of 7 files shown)


supakeen (approved):
Nice; let's get this in as it's a core thing `image-builder` should be able to do so we can then iterate on it.

I see no blockers; code-wise.

internal/testutil/testutil.go
```diff
@@ -0,0 +1,62 @@
...
+	require.NoError(t, err)
+
+	return mockCmd
+}
+
+func (mc *MockCmd) Path() string {
+	return filepath.Join(mc.binDir, mc.name)
+}
+
+func (mc *MockCmd) Restore() error {
```
ondrejbudai, line 41: `t.TempDir` gets removed automatically, so we should not need `Restore` at all. See https://pkg.go.dev/testing#T.TempDir

ondrejbudai (approved):
Awesome, thanks! :)

internal/testutil/testutil.go
mvo5: Thanks, good point! The way we use it currently I guess we could remove it (I'm tempted :) - in the general case it might be useful though if one would want to "unmock" a command during test runs (which is a rare thing to do I guess).

PR #9, opened 2024-12-12
cmd: add new `describe-image` command

mvo5:
[While this should follow https://github.com/osbuild/images/pull/1104, i.e. this output should be used as input in https://github.com/osbuild/images/pull/1104 we could still make it available as an unstable feature?]

This commit adds a new `describe-image` comamnd that contains
the details about the given image type. The output is yaml as
it is both nicely human readable and also machine readable.

The output looks like this:
```yaml
$ ./image-builder describe-image rhel-9.1 tar
   @WARNING - the output format is not stable yet and may change
distro: rhel-9.1
type: tar
arch: x86_64
os_vesion: "9.1"
bootmode: none
partition_type: ""
default_filename: root.tar.xz
packages:
  include:
    - policycoreutils
    - selinux-policy-targeted
    - selinux-policy-targeted
  exclude:
    - rng-tools
```

Thanks to Ondrej Budai for the idea and the example.

### DIFF
--- a/cmd/image-builder/describeimg.go
+++ b/cmd/image-builder/describeimg.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"slices"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/imagefilter"
+)
+
+// Use yaml output by default because it is both nicely human and
+// machine readable and parts of our image defintions will be written
+// in yaml too.  This means this should be a possible input a
+// "flattended" image definiton.
+type describeImgYAML struct {
+	Distro string `yaml:"distro"`
+	Type   string `yaml:"type"`
+	Arch   string `yaml:"arch"`
+
+	// XXX: think about ordering (as this is what the user will see)
+	OsVersion string `yaml:"os_vesion"`
+
+	Bootmode        string `yaml:"bootmode"`
+	PartitionType   string `yaml:"partition_type"`
+	DefaultFilename string `yaml:"default_filename"`
+
+	BuildPipelines   []string                 `yaml:"build_pipelines"`
+	PayloadPipelines []string                 `yaml:"payload_pipelines"`
+	Packages         map[string]*packagesYAML `yaml:"packages"`
+}
+
+type packagesYAML struct {
+	Include []string `yaml:"include"`
+	Exclude []string `yaml:"exclude"`
+}
+
+func packageSetsFor(imgType distro.ImageType) (map[string]*packagesYAML, error) {
+	var bp blueprint.Blueprint
+	manifest, _, err := imgType.Manifest(&bp, distro.ImageOptions{}, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	res := make(map[string]*packagesYAML)
+
+	for pipelineName, pkgSets := range manifest.GetPackageSetChains() {
+		incM := map[string]bool{}
+		excM := map[string]bool{}
+		for _, pkgSet := range pkgSets {
+			for _, s := range pkgSet.Include {
+				incM[s] = true
+			}
+			for _, s := range pkgSet.Exclude {
+				excM[s] = true
+			}
+		}
+		inc := make([]string, 0, len(incM))
+		exc := make([]string, 0, len(excM))
+		for name := range incM {
+			inc = append(inc, name)
+		}
+		for name := range excM {
+			exc = append(exc, name)
+		}
+		slices.Sort(inc)
+		slices.Sort(exc)
+
+		res[pipelineName] = &packagesYAML{
+			Include: inc,
+			Exclude: exc,
+		}
+	}
+	return res, nil
+}
+
+// XXX: should this live in images instead?
+func describeImage(img *imagefilter.Result, out io.Writer) error {
+	// see
+	// https://github.com/osbuild/images/pull/1019#discussion_r1832376568
+	// for what is available on an image (without depsolve or partitioning)
+	pkgSets, err := packageSetsFor(img.ImgType)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


supakeen:
Since #6 has landed for a while now we can probably undraft this.

One thing, perhaps we can call it `describe`? All commands work with images.

mvo5:
While looking at https://github.com/osbuild/bootc-image-builder/pull/808 I was wondering if we could make it easier to figure out what packages are part of an image and then I did remember this PR. I think the output needs work and it ideally would be an input to https://github.com/osbuild/images/pull/1104 but we are not there yet.

So I was thinking in the meantime we simply could have a warning and mark it as an unstable feature (and deliberately break the yaml from being parseable) and then it would be a useful debug tool already. Wdyt?

cmd/image-builder/main.go
```diff
@@ -246,6 +272,20 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	buildCmd.Flags().String("cache", "/var/cache/image-builder/store", `osbuild directory to cache intermediate build artifacts"`)
 	rootCmd.AddCommand(buildCmd)
 
+	// XXX: add --format=json too?
+	describeImgCmd := &cobra.Command{
+		Use:          "describe-image <distro> <image-type>",
+		Short:        "describe the given distro/image-type, e.g. centos-9 qcow2",
+		RunE:         cmdDescribeImg,
+		SilenceUsage: true,
+		Args:         cobra.ExactArgs(1),
```
achilleas-k, line 281: This should be 2.

achilleas-k (changes requested):

cmd/image-builder/main.go
```diff
@@ -246,6 +272,20 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	buildCmd.Flags().String("cache", "/var/cache/image-builder/store", `osbuild directory to cache intermediate build artifacts"`)
 	rootCmd.AddCommand(buildCmd)
 
+	// XXX: add --format=json too?
+	describeImgCmd := &cobra.Command{
+		Use:          "describe-image <distro> <image-type>",
```
achilleas-k, line 277:
```suggestion
		Use:          "describe-image <image-type>",
```
Distro is an optional arg, right?

achilleas-k (changes requested):

cmd/image-builder/main.go
achilleas-k: My mistake.  It's just the use line and example that are wrong.

cmd/image-builder/describeimg.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// XXX: think about ordering (as this is what the user will see)
+	OsVersion string `yaml:"os_vesion"`
+
+	Bootmode        string `yaml:"bootmode"`
+	PartitionType   string `yaml:"partition_type"`
+	DefaultFilename string `yaml:"default_filename"`
+
+	// XXX: add pipelines here? maybe at least exports?
+	Packages *packagesYAML `yaml:"packages"`
```
thozza, line 32: I think the pipeline should be added here, since otherwise the output won't be complete for installer images. It may make sense to do it in a smart way - include pipeline only if here are multiple pipelines with a package set defined? :thinking:

cmd/image-builder/describeimg.go
```diff
@@ -0,0 +1,104 @@
...
+	var bp blueprint.Blueprint
+	manifest, _, err := imgType.Manifest(&bp, distro.ImageOptions{}, nil, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	incM := map[string]bool{}
+	excM := map[string]bool{}
+	// XXX: or should this just do what osbuild-package-sets does
+	// and inlcude what pipeline needs the package set too?
```
thozza, line 50: Personally, I think that merging pipeline package sets makes little sense because it will be confusing and impossible to determine what will be in the "payload" and what will be installed, e.g., in the Anaconda ISO for installer images.

cmd/image-builder/main.go
mvo5: Thanks, fixed now.

cmd/image-builder/describeimg.go
mvo5: Makes sense, I updated the commit to include the pipelines, I hope this captures what you wrote.

thozza:
@mvo5 I randomly picked one failed RPM build in COPR, and it seems to be failing on unit tests, which is strange because they passed in the GH Action. Could you please take a look?

cmd/image-builder/main.go
```diff
@@ -246,6 +272,20 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	buildCmd.Flags().String("cache", "/var/cache/image-builder/store", `osbuild directory to cache intermediate build artifacts"`)
 	rootCmd.AddCommand(buildCmd)
 
+	// XXX: add --format=json too?
```
thozza, line 275: I had the same thought and it would mean basically stripping the `YAML` from struct names such as `describeImgYAML` and adding tags for `json. I think that it would make sense and should be fairly easy to do. One can then use `jq` to extract whatever information they need and things become interesting :sweat_smile:

thozza (changes requested):
LGTM, but unit tests are failing in RPM builds in COPR, so blocking the PR, because those are probably not required.

cmd/image-builder/main.go
mvo5: Thanks, yeah, that makes sense.

I updated the PR, the test was failing because I forgot to pass --arch in the describe-image smoke test so on non-x86 it would fail. this is fixed now.

thozza (approved):

supakeen (approved):

supakeen:
@thozza you have requested changes still :)

mvo5:
@supakeen its actually @achilleas-k how still has a "change-request" open (the GH ui is a bit confusing)

supakeen:
Sorry!

achilleas-k (approved):

PR #11, opened 2024-12-13
misc: add Containerfile

mvo5:
This PR adds Containerfile integration, with that a bib style deploy/run is possible. The Containerfile is pretty nasty right now, especially around the "getting supported repositories" part. This needs to get fixed either in composer or images IMHO (or via go:embed). But with that we can give it to our early users and let them play with building images.

It also contains automatic container building as part of the GH action (thanks @supakeen!) and once we have it we can update the README to point to the auto-generated containers for even easier testing/using.

As is "list-images" and "manifest" will work but once #8 is merged the below works (on my local system, we want tests here :)

Run build via container, bib-style:
```console
$ sudo podman build . -t test-icbli
$ sudo podman run --privileged \
   -v ./output:/output \
   test-ibcli:latest \
   build \
   --distro fedora-41 \
   minimal-raw
```

### DIFF
--- a/.github/workflows/container.yaml
+++ b/.github/workflows/container.yaml
@@ -1,0 +1,48 @@
+name: Build & Deploy Container
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  merge_group:
+    types: [ "checks_requested" ]
+
+jobs:
+  build-base:
+    name: Build & push image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Github Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Generate container image metadata
+        uses: docker/metadata-action@v4
+        id: meta
+        with:
+          images: ghcr.io/osbuild/image-builder-cli
+          tags: |
+            type=ref,event=tag,enable=${{ startsWith(github.ref, 'refs/tags/') }}
+            type=raw,value=latest
+            type=sha,format=long,enable=${{ github.ref == 'refs/heads/main' }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          platforms: linux/amd64
+          file: ./Containerfile
+          push: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
+          cache-from: type=registry,ref=ghcr.io/osbuild/image-builder-cli:latest
+          cache-to: type=inline
+          tags: ${{ steps.meta.outputs.tags }}
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -13,6 +13,7 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
+    uses:  ./.github/workflows/testdeps.yml
     steps:
     - uses: actions/checkout@v4
 
@@ -21,13 +22,6 @@ jobs:
       with:
         go-version: 'stable'
         
-    - name: Apt update
-      run: sudo apt update
-
-    # This is needed for the container resolver dependencies
-    - name: Install test dependencies
-      run: sudo apt install -y libgpgme-dev libbtrfs-dev libdevmapper-dev podman
-
     - name: Build
       run: go build -v ./...
 
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,24 @@
+name: Integration tests
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    uses:  ./.github/workflows/testdeps.yml
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install integration test env
+      run: |
+        sudo apt update
+        sudo apt install -y pytest golang
+
+    - name: Run integration tests via pytest
+      run: |
+        # use "-s" for now for easier debugging
+        sudo pytest -s -v
--- a/Containerfile
+++ b/Containerfile
@@ -48,8 +48,8 @@ COPY entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
 VOLUME /output
 WORKDIR /output
+# XXX: add "store" flag like bib
 VOLUME /store
-VOLUME /rpmmd
 VOLUME /var/lib/containers/storage
 
 LABEL description="This tools allows to build and deploy disk-images."
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,59 @@
+FROM registry.fedoraproject.org/fedora:41 AS builder
+RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/
+ARG GOPROXY=https://proxy.golang.org,direct
+RUN go env -w GOPROXY=$GOPROXY
+COPY . /build
+WORKDIR /build
+# keep in sync with:
+# https://github.com/containers/podman/blob/2981262215f563461d449b9841741339f4d9a894/Makefile#L51
+RUN go build -tags "containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper" ./cmd/image-builder
+
+FROM registry.fedoraproject.org/fedora:41
+
+# Fast-track osbuild so we don't depend on the "slow" Fedora release process to implement new features in bib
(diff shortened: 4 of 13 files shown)


supakeen:
I've been rude and committed CI for the container build, we might need to change organisation or repository permissions for it to upload.

supakeen (approved):
Looks good (though I guess I wrote part of it so maybe wait for a second review).

Containerfile
```diff
@@ -0,0 +1,59 @@
...
+RUN go env -w GOPROXY=$GOPROXY
+COPY . /build
+WORKDIR /build
+# keep in sync with:
+# https://github.com/containers/podman/blob/2981262215f563461d449b9841741339f4d9a894/Makefile#L51
+RUN go build -tags "containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper" ./cmd/image-builder
+
+FROM registry.fedoraproject.org/fedora:41
+# Fast-track osbuild so we don't depend on the "slow" Fedora release process to implement new features in bib
+RUN dnf install -y dnf-plugins-core
```
ondrejbudai, line 13: Can we put these in a single RUN command, and clean the dnf cache afterwards, to save some disk space and reduce the number of layers?

```
RUN dnf install -y dnf-plugins-core \
  && dnf copr enable -y @osbuild/osbuild \
  && dnf install -y libxcrypt-compat wget osbuild osbuild-ostree osbuild-depsolve-dnf osbuild-lvm2 \
  && dnf clean all
```

README.md
```diff
@@ -2,6 +2,19 @@
 
 Build images from the commandline in a convenient way.
 
+## Run via container
+
+TODO: build container automatically via GH action and put into ghcr.io
```
ondrejbudai, line 7: Leftover TODO

entrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+set -e
+
+# TODO: share code with bib to do the setup automatically
+# see https://github.com/teamsbc/container-for-osbuild/blob/main/entrypoint.bash (thanks simon)
+# and https://github.com/osbuild/bootc-image-builder/blob/main/bib/internal/setup/setup.go#L21 (thanks ondrej,achilleas,colin)
+mkdir /run/osbuild
+mkdir /run/osbuild-store
+
+mount -t tmpfs tmpfs /run/osbuild
+mount -t tmpfs tmpfs /run/osbuild-store
```
ondrejbudai, line 12: I think this is currently unused, and I think we should actually use a "real" filesystem for the osbuild store, because tmpfs might not be big enough for image building.

Containerfile
```diff
@@ -0,0 +1,59 @@
...
+ cp -a centos-stream-9.json centos-9.json
+ cp -a centos-stream-10.json centos-10.json
+EOR
+
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+VOLUME /output
+WORKDIR /output
+VOLUME /store
+VOLUME /rpmmd
```
ondrejbudai, line 52: I think this is currently unused.

Containerfile
```diff
@@ -0,0 +1,59 @@
...
+ # XXX: find an even better way here, those are symlinks in the upstream repo
+ cp -a centos-stream-9.json centos-9.json
+ cp -a centos-stream-10.json centos-10.json
+EOR
+
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+VOLUME /output
+WORKDIR /output
+VOLUME /store
```
ondrejbudai, line 51: This is currently unused, and should be used (osbuild's store cannot be on on overlayfs). How can we pass the `/store` path to image builder cli to utilize it for the osbuild's store?

.github/workflows/container.yaml
```diff
@@ -0,0 +1,46 @@
+name: Build & Deploy Container
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
```
ondrejbudai, line 6: Also build in a merge queue?

Containerfile
supakeen: I've addressed this in a separate commit.

README.md
supakeen: I've addressed this in a separate commit.

supakeen:
I've made 3 separate commits to take care of the low hanging fruit things, feel free to reorganize/rebase how you see fit @mvo5 

(I don't want to force push to your branch at all).

The other remarks will take a bit more thinking.

entrypoint.sh
mvo5: Thanks, indeed, I removed this and followed our pattern from bib and added `--store` and use that in the entrypoint.

Containerfile
mvo5: Indeed, I removed it (for now, I think we want to add this later again once we grow a --rpmmd like flag).

mvo5:
> I've made 3 separate commits to take care of the low hanging fruit things, feel free to reorganize/rebase how you see fit @mvo5
> 
> (I don't want to force push to your branch at all).
> 
> The other remarks will take a bit more thinking.

Thank you! This is really appreciated! I squashed them into their relevant places (hopefully :) and also addressed the points from @ondrejbudai - I think this is ready for a re-review. I can split out further if needed but it seems it's not toooo big .

achilleas-k (approved):
LGTM.

achilleas-k:
One thought about the new `--store` flag (to consider in a follow-up):
Would it make sense to separate osbuild flags into their own group?  Could we have an `--osbuild-opts` flag for example that can take options to be passed down to osbuild?

So we could do:
```
image-builder-cli ... --osbuild-opts="store=/data/osbuild-store,checkpoint=build,checkpoint=os,cache-max-size='2 GiB'" ...
```

It's not the prettiest but it saves us from having to support every osbuild flag.  OTOH, it also forces users to look up osbuild flags if they want to discover everything that's supported.

If we don't like this, I think eventually we should support every option that osbuild supports (except maybe the `--monitor` option).

.github/workflows/go.yml
```diff
@@ -13,6 +13,7 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
+    uses:  ./.github/workflows/testdeps.yml
```
ondrejbudai, line 16: 😍

test/containerbuild.py
```diff
@@ -0,0 +1,23 @@
+import os
```
ondrejbudai, line 1: I think we just need `subprocess` and `pytest` here.

test/test_container.py
```diff
@@ -0,0 +1,27 @@
+import os
```
ondrejbudai, line 1: Would you be willing to use Go for integration tests? Is there any reason to actually use Python?

I can tell you why I prefer Go:
- developers don't need to switch their language mindset when writing integration tests
- we don't need an extra linter and formatter
- sharing code between integration tests and the SUT
- pytest fixtures are cool... until they start being overwhelming. I think I have a preference for the lightweight nature of Go.

Containerfile
```diff
@@ -0,0 +1,59 @@
...
+ cp -a centos-stream-9.json centos-9.json
+ cp -a centos-stream-10.json centos-10.json
+EOR
+
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+VOLUME /output
+WORKDIR /output
+# XXX: add "store" flag like bib
+VOLUME /store
```
ondrejbudai, line 52: I had an idea: image-builder-cli is supposed to be an RPM in the future. So this path should be FHS-compliant, and ideally it should be the same between the container image and the RPM.

Something like `/var/cache/image-builder/{store,dnf-cache}`?

Note that I have no concerns about `/output`, this is perfectly valid to be container-specific.

test/test_container.py
```diff
@@ -0,0 +1,27 @@
...
+    output_dir = tmp_path / "output"
+    output_dir.mkdir()
+    subprocess.check_call([
+        "podman", "run",
+        "--privileged",
+        "-v", f"{output_dir}:/output",
+        build_container,
+        "build",
+        "minimal-raw",
+        "--distro", "fedora-41"
```
ondrejbudai, line 20: Can we do something quick and dirty to pave over the RPM downloading issue that our stack is suffering from? Override repos with rpmrepo.osbuild.org for the test?

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,24 @@
+name: Integration tests
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
ondrejbudai, line 7: no merge queue here

Containerfile
mvo5: Love it! will do in a followup

test/test_container.py
mvo5: I will try centos-9 instead, from my experience with bib it seems the centos mirrors are generally a bit more stable.

test/containerbuild.py
mvo5: Yeah, silly me - and a linter :)

test/test_container.py
mvo5: Sure, I will give this a "go" (see what I did here ;)

PR #14, opened 2024-12-18
package: specfile

supakeen:
Provides a simplified version of the spec file and makefile included in `osbuild-composer`. Vendors existing dependencies, and enables CI (dependabot) to bump those.


mvo5 (commented):
A high level question/suggestion - can we just not include the vendor/ dir but instead make downloading/creating it part of the "make rpm" makefile? I would really like to try to avoid adding 1 million LoC if we can :)

supakeen:
> A high level question/suggestion - can we just not include the vendor/ dir but instead make downloading/creating it part of the "make rpm" makefile? I would really like to try to avoid adding 1 million LoC if we can :)

The vendor directory needs to be part of our 'source distribution' which currently is a git checkout :)

supakeen:
Close in favor of #15.

PR #15, opened 2024-12-18
package: specfile

schuellerf:
This is a maybe controversial version of #14 which tries to avoid commiting `vendor/` but patching source tar ball.
Prerequisite to testing https://github.com/osbuild/release-action/pull/25

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -17,7 +17,7 @@ srpm_build_deps:
 actions:
   # v1 just the default prior to having tags
   get-current-version: bash -c "( git describe --tags --abbrev=0 2>/dev/null || echo v1 ) | sed 's|v||'"
-  post-upstream-clone: bash -c "go mod vendor || ./tools/rpm_spec_add_provides_bundle.sh"
+  post-upstream-clone: bash -c "go mod vendor && ./tools/rpm_spec_add_provides_bundle.sh"
   create-archive: bash -c "make release_artifacts"
 
 # Handle only releases without a "dot" (e.g. v88.2), since "dot" releases should never be released to Fedora
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -15,7 +15,8 @@ srpm_build_deps:
   - golang
 
 actions:
-  get-current-version: bash -c "git describe --tags --abbrev=0 | sed 's|v||' && echo 1"
+  # v1 just the default prior to having tags
+  get-current-version: bash -c "( git describe --tags --abbrev=0 2>/dev/null || echo v1 ) | sed 's|v||'"
   post-upstream-clone: bash -c "go mod vendor || ./tools/rpm_spec_add_provides_bundle.sh"
   create-archive: bash -c "make release_artifacts"
 
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -12,10 +12,11 @@ upstream_tag_template: v{version}
 
 srpm_build_deps:
   - make
+  - golang
 
 actions:
   get-current-version: bash -c "git describe --tags --abbrev=0 | sed 's|v||'"
-  post-upstream-clone: "./tools/rpm_spec_add_provides_bundle.sh"
+  post-upstream-clone: bash -c "go mod vendor && ./tools/rpm_spec_add_provides_bundle.sh"
   create-archive: bash -c "make release_artifacts"
 
 # Handle only releases without a "dot" (e.g. v88.2), since "dot" releases should never be released to Fedora
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,60 @@
+# https://packit.dev/docs/configuration/
+
+specfile_path: image-builder-cli.spec
+
+files_to_sync:
+    - image-builder-cli.spec
+    - .packit.yaml
+
+copy_upstream_release_description: true
+
+upstream_tag_template: v{version}
+
+srpm_build_deps:
+  - make
+
+actions:
+  get-current-version: bash -c "git describe --tags --abbrev=0 | sed 's|v||'"
+  post-upstream-clone: "./tools/rpm_spec_add_provides_bundle.sh"
+  create-archive: bash -c "make release_artifacts"
+
+# Handle only releases without a "dot" (e.g. v88.2), since "dot" releases should never be released to Fedora
+upstream_tag_include: 'v\d+'
+
+jobs:
+# image-builder-cli is not yet in Fedora, so we don't need to update it there
+# - job: bodhi_update
+#   trigger: commit
+#   dist_git_branches:
+#     - fedora-branched # rawhide updates are created automatically
+# - job: koji_build
+#   trigger: commit
+#   dist_git_branches:
+#     - fedora-all
+# - job: propose_downstream
+#   trigger: release
+#   dist_git_branches:
+#     - fedora-all
+- job: copr_build
+  trigger: pull_request
+  targets: &build_targets
+    - centos-stream-9-aarch64
+    - centos-stream-9-s390x
+    - centos-stream-9-ppc64le
+    - centos-stream-9-x86_64
+    - centos-stream-10-aarch64
+    - centos-stream-10-s390x
+    - centos-stream-10-ppc64le
+    - centos-stream-10-x86_64
+    - fedora-all-aarch64
+    - fedora-all-s390x
+    - fedora-all-ppc64le
+    - fedora-all
+    - rhel-9-aarch64
+    - rhel-9-x86_64
+- job: copr_build
+  trigger: commit
+  branch: main
+  owner: "@osbuild" # copr repo namespace
+  project: image-builder-cli  # copr repo name so you can consume the builds
+  targets: *build_targets
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -15,8 +15,8 @@ srpm_build_deps:
   - golang
 
 actions:
-  get-current-version: bash -c "git describe --tags --abbrev=0 | sed 's|v||'"
-  post-upstream-clone: bash -c "go mod vendor && ./tools/rpm_spec_add_provides_bundle.sh"
+  get-current-version: bash -c "git describe --tags --abbrev=0 | sed 's|v||' && echo 1"
+  post-upstream-clone: bash -c "go mod vendor || ./tools/rpm_spec_add_provides_bundle.sh"
   create-archive: bash -c "make release_artifacts"
 
 # Handle only releases without a "dot" (e.g. v88.2), since "dot" releases should never be released to Fedora
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,158 @@
+#
+# Maintenance Helpers
+#
+# This makefile contains targets used for development, as well as helpers to
+# aid automatization of maintenance. Unless a target is documented in
+# `make help`, it is not supported and is only meant to be used by developers
+# to aid their daily development work.
+#
+# All supported targets honor the `SRCDIR` variable to find the source-tree.
+# For most unsupported targets, you are expected to have the source-tree as
+# your working directory. To specify a different source-tree, simply override
+# the variable via `SRCDIR=<path>` on the commandline. By default, the working
+# directory is used for build output, but `BUILDDIR=<path>` allows overriding
+# it.
+#
+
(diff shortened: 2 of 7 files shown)


mvo5 (commented):
Thank you! I really like this approach!

thozza:
I'm sorry, but as far as I can tell, this won't work for releasing to Fedora, CS, and RHEL.

Once we start using GitHub action to push release tags, GH will generate the tarball for us, and it will miss the vendored packages. This PR solves only RPM builds in CI using Makefile using a full repo checkout, which won't work with downstream repositories. The problem that needs to be solved is that the release tarball attached to GH releases needs to contain all vendored code.

supakeen:
> I'm sorry, but as far as I can tell, this won't work for releasing to Fedora, CS, and RHEL.
> 
> Once we start using GitHub action to push release tags, GH will generate the tarball for us, and it will miss the vendored packages. This PR solves only RPM builds in CI using Makefile using a full repo checkout, which won't work with downstream repositories. The problem that needs to be solved is that the release tarball attached to GH releases needs to contain all vendored code.
> 
> EDIT: OK, with the modified release action, this should work. Looking forward to see this work in action 🤔

I'm still not fully on-board as I've had to deal with people excluding/including things in release tarballs way too often while packaging stuff. I'm not entirely up to date on the Go macros but do they use the actual release tarballs or do they check out the tag from git?

We'd also need to run `make` in PackIt right?

I prefer that a certain `git` checkout is always buildable.

Makefile
```diff
@@ -0,0 +1,173 @@
...
+$(RPM_SPECFILE):
+	mkdir -p $(CURDIR)/rpmbuild/SPECS
+	git show HEAD:image-builder-cli.spec > $(RPM_SPECFILE)
+	./tools/rpm_spec_add_provides_bundle.sh $(RPM_SPECFILE)
+
+RPM_TARBALL_UNCOMPRESSED=$(RPM_TARBALL:.tar.gz=.tar)
+
+$(RPM_TARBALL):
+	mkdir -p $(CURDIR)/rpmbuild/SOURCES
+	git archive --prefix=image-builder-cli-$(COMMIT)/ --format=tar.gz HEAD > $(RPM_TARBALL)
```
thozza, line 136: This should probably consider release tags `v*` so that our release tarballs have a nice filename that consists only of the name and version but no commit ref.

Makefile
```diff
@@ -0,0 +1,173 @@
...
+
+.PHONY: $(RPM_SPECFILE)
+$(RPM_SPECFILE):
+	mkdir -p $(CURDIR)/rpmbuild/SPECS
+	git show HEAD:image-builder-cli.spec > $(RPM_SPECFILE)
+	./tools/rpm_spec_add_provides_bundle.sh $(RPM_SPECFILE)
+
+RPM_TARBALL_UNCOMPRESSED=$(RPM_TARBALL:.tar.gz=.tar)
+
+$(RPM_TARBALL):
```
thozza, line 134: The tarball should ontain SPEC file with proper `Provides` for bundled code, so the commands executed in the `$(RPM_SPECFILE)` target should be executed also on the SPEC that ends up in the release tarball.

Makefile
```diff
@@ -0,0 +1,173 @@
...
+# commit, based on the spec file in this directory.  The resulting packages
+# have the commit hash in their version, so that they don't get overwritten
+# when calling `make rpm` again after switching to another branch.
+#
+# All resulting files (spec files, source rpms, rpms) are written into
+# ./rpmbuild, using rpmbuild's usual directory structure.
+#
+
+RPM_SPECFILE=rpmbuild/SPECS/image-builder-cli.spec
+RPM_TARBALL=rpmbuild/SOURCES/image-builder-cli-$(COMMIT).tar.gz
```
thozza, line 124: Tarball should IMO contain also the release version, not just the commit.

schuellerf:
@supakeen 
> We'd also need to run `make` in PackIt right?
Not necessarily, the action will take care, that the release contains the tarball with vendored stuff already.
( see https://github.com/osbuild/release-action/blob/staging/action.yml#L46 )

> I prefer that a certain `git` checkout is always buildable.
It is - `go.mod` contains the versions anyway and will fetch when building…
So either fetch the source tarball with vendor stuff or just build with go

Makefile
schuellerf: Thanks for the hint!

Makefile
```diff
@@ -0,0 +1,173 @@
...
+#         be set to the latest version number of this repository without any
+#         prefix (just a plain number).
+#
+#     COMMIT:
+#         This evaluates to the latest git commit sha. This will not work if
+#         the source is not a git checkout. Hence, this variable is not
+#         pre-fetched but evaluated at time of use.
+#
+
+VERSION := $(shell (cd "$(SRCDIR)" && grep "^Version:" image-builder-cli.spec | sed 's/[^[:digit:]]*\([[:digit:]]\+\).*/\1/'))
```
thozza, line 50: This should take potential "dot" versions into consideration. Such as `12.1`, since we use that for backporting. Moreover, this variable does not seem to be used anywhere at all.

Makefile
schuellerf: I'll align with @mvo5 and @supakeen

schuellerf:
> EDIT: OK, with the modified release action, this should work. Looking forward to see this work in action 🤔

me too 😄🎉

schuellerf:
> PackIt builds from commits (and releases) but the `vendor` directory would not be there in commits?

That's the point of this approach, that the release contains the vendor directory.

> That works well when you have network access in your build environment. While I have that on my own workstation I don't have it everywhere.

So you'll just download the source tarball _with_ vendored stuff

supakeen:
> > PackIt builds from commits (and releases) but the `vendor` directory would not be there in commits?
> 
> That's the point of this approach, that the release contains the vendor directory.
> 
> > That works well when you have network access in your build environment. While I have that on my own workstation I don't have it everywhere.
> 
> So you'll just download the source tarball _with_ vendored stuff

Which doesn't exist for each commit, only tagged releases?

Makefile
schuellerf: I'll fix it - but it's wrong here too
https://github.com/osbuild/osbuild-composer/releases/tag/v127

thozza:
> > > PackIt builds from commits (and releases) but the `vendor` directory would not be there in commits?
> > 
> > 
> > That's the point of this approach, that the release contains the vendor directory.
> > > That works well when you have network access in your build environment. While I have that on my own workstation I don't have it everywhere.
> > 
> > 
> > So you'll just download the source tarball _with_ vendored stuff
> 
> Which doesn't exist for each commit, only tagged releases?

The answer is YES, Packit will need to have a custom action to build the archive with vendored code for PRs.

.packit.yaml
```diff
@@ -12,10 +12,11 @@ upstream_tag_template: v{version}
 
 srpm_build_deps:
   - make
+  - golang
 
 actions:
   get-current-version: bash -c "git describe --tags --abbrev=0 | sed 's|v||'"
-  post-upstream-clone: "./tools/rpm_spec_add_provides_bundle.sh"
+  post-upstream-clone: bash -c "go mod vendor && ./tools/rpm_spec_add_provides_bundle.sh"
```
mvo5, line 19: :heart:

supakeen:
Let's undraft this one, I've closed #14 as efforts seem to be more focused on this PR and while I have my own opinions it's not really important enough if everyone else doesn't want `/vendor`.

I'd really like if we have RPMs shipping today :)

image-builder-cli.spec
```diff
@@ -0,0 +1,97 @@
...
+Version:        0
+
+%gometa
+
+%global common_description %{expand:
+A service for building customized OS artifacts, such as VM images and OSTree
+commits, that uses osbuild under the hood. Besides building images for local
+usage, it can also upload images directly to cloud.
+
+It is compatible with composer-cli and cockpit-composer clients.
```
thozza, line 20: This was probably not changed when copied over from osbuild-composer. We should update it.

image-builder-cli.spec
```diff
@@ -0,0 +1,97 @@
...
+# Upstream license specification: Apache-2.0
+License:        Apache-2.0
+URL:            %{gourl}
+Source0:        %{gosource}
+
+
+BuildRequires:  %{?go_compiler:compiler(go-compiler)}%{!?go_compiler:golang}
+BuildRequires:  systemd
+BuildRequires:  krb5-devel
+BuildRequires:  python3-docutils
```
thozza, line 37: I'm not sure if they are still relevant, because we do not interact with systemd or Koji (krb5) and I'm not sure about docs...

image-builder-cli.spec
```diff
@@ -0,0 +1,97 @@
...
+BuildRequires:  make
+# Build requirements of 'theproglottis/gpgme' package
+BuildRequires:  gpgme-devel
+BuildRequires:  libassuan-devel
+# Build requirements of 'github.com/containers/storage' package
+BuildRequires:  device-mapper-devel
+%if 0%{?fedora}
+BuildRequires:  systemd-rpm-macros
```
thozza, line 45: This is, in my opinion, not needed because we do not enable/disable/start/stop/restart any services as part of the RPM installation process.

image-builder-cli.spec
supakeen: These are likely leftovers from the copy :)

image-builder-cli.spec
supakeen: These are likely leftovers from the copy :)

thozza:
@schuellerf the sources archive should really be named `image-builder-cli-${version_from_spec}.tar.gz`. This is what the `rpmbuild` tooling expects and this is how things work on any of our other repositories.

see https://download.copr.fedorainfracloud.org/results/packit/osbuild-image-builder-cli-15/fedora-41-x86_64/08414991-image-builder-cli/builder-live.log.gz
```
+ /usr/lib/rpm/rpmuncompress -x /builddir/build/SOURCES/image-builder-cli-65b31cf96d13561bca0b30f455e48874ead1e196.tar.gz
+ STATUS=0
+ '[' 0 -ne 0 ']'
+ cd image-builder-cli-1
/var/tmp/rpm-tmp.vardlL: line 41: cd: image-builder-cli-1: No such file or directory
```

Makefile
schuellerf: @mvo5 @supakeen @thozza should I just manually set the tag `v1` for now?

Makefile
thozza: Yes, it is a bug in our osbuild-composer tarball. We do not have proper provides for bundled code in CS/RHEL. :-/

schuellerf:
@thozza 
> @schuellerf the sources archive should really be named `image-builder-cli-${version_from_spec}.tar.gz`. This is what the `rpmbuild` tooling expects and this is how things work on any of our other repositories.

Sound really reasonable, but why would this work with `osbuild-composer` then :-/
https://github.com/osbuild/osbuild-composer/blob/main/Makefile#L317

thozza:
So, https://github.com/osbuild/image-builder-cli/pull/15/commits/33d65c2de88e2fb1c0ea25861a5d3fdeb4171e1d fixed build in COPR (and would make things work in downstream distros), but it broke the `make rpm` use case. In COPR and build systems, we can't define `commit` when building (S)RPM. So we need to build the tarball conditionally with `$(VERSION)` or `$(COMMIT)` based on the use case.

thozza:
> @thozza
> 
> > @schuellerf the sources archive should really be named `image-builder-cli-${version_from_spec}.tar.gz`. This is what the `rpmbuild` tooling expects and this is how things work on any of our other repositories.
> 
> Sound really reasonable, but why would this work with `osbuild-composer` then :-/ https://github.com/osbuild/osbuild-composer/blob/main/Makefile#L317

This would work only for CI or local invocations of `rpmbuild` with the `commit` variable defined. We can't do that in COPR and downstream build systems.

ondrejbudai:
For the record, cockpit does something similar for node_modules, so it's a viable method. :)

See:
- https://github.com/cockpit-project/cockpit/blob/main/packit.yaml#L3
- the extra archive at https://github.com/cockpit-project/cockpit/releases/tag/331

supakeen (approved):

PR #17, opened 2024-12-19
ibcli: use `/var/cache/image-builder/store` as default for `--store`

mvo5:
This commit follows a suggestion from Ondrej to add a better default for `--store`: /var/cache/image-builder/store.

This is nice as it will give users automatic caching and makes the container also nicer.


ondrejbudai (approved):
Awesome, thanks!

PR #18, opened 2024-12-19
test: move to `centos-9` for the smoke test

mvo5:
This commit moves the smoke test from fedora-41 to centos-9. The reason is that it seems the fedora mirrors in general are a bit more flaky so /hopefully/ the centos one will help.

The real fix of course is to retry different mirrors on failure, we will need to look into librepo integration in osbuild (again) for this (hopefully soon!).


ondrejbudai (approved):
Makes sense :)

PR #20, opened 2024-12-19
workflow: rename all `.yml` files to `.yaml`

mvo5:
The original `go.yml` got added via some GH automation. But we are not MSDOS and we can affort more than 3 chars for a file extension - so this commit renames all .yml to .yaml which is what we use everywhere else.


supakeen (approved):

supakeen:
> we are not MSDOS

(citation needed)

ondrejbudai (approved):
lol

PR #21, opened 2024-12-19
main: add --progres (similar to bib)

mvo5:
[draft as this lacks tests,  not easy to do right now without duplicating code from bib, looking into this]

This commit add progress reporting similar to what `bib` is doing.
It imports the progress from `bootc-image-builder` for now. There
is an open PR for moving this to images but there are some
concerns about moving it there so for now we just use bib.
This is not too bad because eventually bib and ibcli will merge.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,11 +2,14 @@ module github.com/osbuild/image-builder-cli
 
 // keep in sync with images,
 // c.f. https://github.com/osbuild/images/blob/main/go.mod
-go 1.21.0
+go 1.22.0
+
+toolchain go1.22.2
 
 require (
 	github.com/BurntSushi/toml v1.4.0
 	github.com/gobwas/glob v0.2.3
+	github.com/osbuild/bootc-image-builder/bib v0.0.0-20250125114430-c5d674173ba0
 	github.com/osbuild/images v0.112.0
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.1
@@ -20,6 +23,7 @@ require (
 	github.com/VividCortex/ewma v1.2.0 // indirect
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d // indirect
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2 // indirect
+	github.com/cheggaaa/pb/v3 v3.1.6 // indirect
 	github.com/containerd/cgroups/v3 v3.0.3 // indirect
 	github.com/containerd/errdefs v0.1.0 // indirect
 	github.com/containerd/stargz-snapshotter/estargz v0.15.1 // indirect
@@ -37,6 +41,7 @@ require (
 	github.com/docker/docker-credential-helpers v0.8.2 // indirect
 	github.com/docker/go-connections v0.5.0 // indirect
 	github.com/docker/go-units v0.5.0 // indirect
+	github.com/fatih/color v1.18.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
 	github.com/go-jose/go-jose/v4 v4.0.2 // indirect
 	github.com/go-logr/logr v1.4.2 // indirect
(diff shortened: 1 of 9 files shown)


supakeen:
Very awesome and works for me; can we get rid of the `WARN[0000] Failed to load consumer certs: no consumer key found`?

supakeen:
SRPM builds are failing due to the merge conflict in `go.sum`; otherwise I think this can be undrafted now that we use `bib`'s code?

mvo5:
> Very awesome and works for me; can we get rid of the `WARN[0000] Failed to load consumer certs: no consumer key found`?

This should fix that particular nit: https://github.com/osbuild/bootc-image-builder/pull/810 

I marked this ready for review now, it contains a sad commit with some duplicated container building code (small but still sad) but this should make it ready and it unblocks the `--with-buildlog` in #106

supakeen:
> > Very awesome and works for me; can we get rid of the `WARN[0000] Failed to load consumer certs: no consumer key found`?
> 
> This should fix that particular nit: [osbuild/bootc-image-builder#810](https://github.com/osbuild/bootc-image-builder/pull/810)
> 
> I marked this ready for review now, it contains a sad commit with some duplicated container building code (small but still sad) but this should make it ready and it unblocks the `--with-buildlog` in #106

I'm not too sad about it, knowing that it's hopefully temporary.

The test is failing due to no longer getting the `error on stderr` bits in the output for `TestBuildIntegration`. This seems related to https://github.com/osbuild/bootc-image-builder/pull/810 ?

mvo5:
> > > Very awesome and works for me; can we get rid of the `WARN[0000] Failed to load consumer certs: no consumer key found`?
> > 
> > 
> > This should fix that particular nit: [osbuild/bootc-image-builder#810](https://github.com/osbuild/bootc-image-builder/pull/810)
> > I marked this ready for review now, it contains a sad commit with some duplicated container building code (small but still sad) but this should make it ready and it unblocks the `--with-buildlog` in #106
> 
> I'm not too sad about it, knowing that it's hopefully temporary. Having the progress reporting is very nice.
> 
> The test is failing due to no longer getting the `error on stderr` bits in the output for `TestBuildIntegration`. This seems related to [osbuild/bootc-image-builder#810](https://github.com/osbuild/bootc-image-builder/pull/810) ?

Thanks, nice catch! I will need to update the test, the facts that we just forwarded stderr from osbuild unfiltered is a bit of a misfeature, I think for progress=term we can/should fix that, for progress=verbose I need to look, I think there should be compatibility code.

go.mod
```diff
@@ -2,11 +2,14 @@ module github.com/osbuild/image-builder-cli
 
 // keep in sync with images,
 // c.f. https://github.com/osbuild/images/blob/main/go.mod
-go 1.21.0
+go 1.22.0
```
mvo5, line 5: This is a bit of an odd one, bib is at 1.22 while images is at 1.21

supakeen: Well; I guess it should stick at the lowest common denominator being 1.21 right?

schuellerf: Didn't we agree on the highest supported by all distros to be 1.22?

schuellerf: (i.e. all of our supported distros ;) )

supakeen: I don't know, if we did then it should be 1.22 in images 😄.

schuellerf: agree - I'm not the right person to verify that 1.22 is supported "everywhere", just vaguely remember this is the way :-D

achilleas-k: 1.22 is in RHEL 9.5 so we're good to go for that.
We can move to 1.23 in images as soon as the IB service container (the UBI) starts using 9.6.

mvo5: Cool, great to hear that this is not a blocker.

mvo5:
> > > > Very awesome and works for me; can we get rid of the `WARN[0000] Failed to load consumer certs: no consumer key found`?
> > > 
> > > 
> > > This should fix that particular nit: [osbuild/bootc-image-builder#810](https://github.com/osbuild/bootc-image-builder/pull/810)
> > > I marked this ready for review now, it contains a sad commit with some duplicated container building code (small but still sad) but this should make it ready and it unblocks the `--with-buildlog` in #106
> > 
> > 
> > I'm not too sad about it, knowing that it's hopefully temporary. Having the progress reporting is very nice.
> > The test is failing due to no longer getting the `error on stderr` bits in the output for `TestBuildIntegration`. This seems related to [osbuild/bootc-image-builder#810](https://github.com/osbuild/bootc-image-builder/pull/810) ?
> 
> Thanks, nice catch! I will need to update the test, the facts that we just forwarded stderr from osbuild unfiltered is a bit of a misfeature, I think for progress=term we can/should fix that, for progress=verbose I need to look, I think there should be compatibility code.

The test is updated now, things are a little bit more complicated than I had hoped but its not too bad, the tests now nicely check for the error handling if osbuild fails which I think is a good thing (tm).

supakeen (approved):
Very cool; some minor tidbits that could be changed but nothing that is a blocker. Seems like the Go version things have been resolved as well so we're ready to go!

PR #29, opened 2024-12-20
github/workflows: introduce release action

schuellerf:
Using staging branch for now, due to the new
release_artifacts.


schuellerf:
I would just merge this and check if it works (as we can't test GH actions otherwise).
The needed secrets should be available implicitly.

mvo5 (approved):
Thank you, this looks great!

PR #30, opened 2024-12-20
image-builder-cli.spec: fix building centos stream 10

schuellerf:
Fix building centos stream 10


mvo5 (commented):
Fwiw, this looks fine to me especially if it builds - but I'm not an rpm expert so a second look might be useful

image-builder-cli.spec
```diff
@@ -63,6 +63,12 @@ BuildRequires:  btrfs-progs-devel
 
 %build
 export GOFLAGS="-buildmode=pie"
+
+# get centos 10 building by enabling modules
+%if 0%{?centos} >= 10
+export GO111MODULE=on
```
supakeen, line 69: This should do the same as we currently do on Fedora. Unsetting `gomodulemode` is enough to get it to work :) I'd also recommend using `?rhel` instead (both are set on CentOS).

thozza: I'm wondering why this is not happening in `osbuild-composer`. We should use a common approach in both projects.

thozza: https://github.com/osbuild/osbuild-composer/blob/e90010f153399e3452884f6ed8da2fcd68d1ffa0/osbuild-composer.spec#L79-L87

achilleas-k (approved):

PR #38, opened 2024-12-20
manifestgen: port to the new `images` manifest seed

mvo5:
This moves to the new images Manifest() API from PR#1107.


supakeen (approved):
Thank you.

PR #51, opened 2025-01-08
main: add `--use-librepo` to support librepo  downloads

mvo5:
With that it then adds a new `--use-librepo` switch that will
enable librepo based downloading so that people can use
the new backend.

It is on by default currently, the risk is that it got no real testing
with secrets (but there code is there and similar to what curl
is doing) but the upside is that it will help with the centos/fedora
cases as there the mirror retry feature is import important.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -199,7 +199,7 @@ operating sytsems like centos and RHEL with easy customizations support.`,
 	manifestCmd.Flags().String("ostree-ref", "", `OSTREE reference`)
 	manifestCmd.Flags().String("ostree-parent", "", `OSTREE parent`)
 	manifestCmd.Flags().String("ostree-url", "", `OSTREE url`)
-	manifestCmd.Flags().Bool("use-librepo", false, `(experimental) use librepo to download packages, needs new osbuild`)
+	manifestCmd.Flags().Bool("use-librepo", true, `use librepo to download packages (disable if you use old versions of osbuild)`)
 	rootCmd.AddCommand(manifestCmd)
 
 	buildCmd := &cobra.Command{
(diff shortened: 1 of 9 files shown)


cmd/image-builder/main.go
```diff
@@ -185,6 +199,7 @@ operating sytsems like centos and RHEL with easy customizations support.`,
 	manifestCmd.Flags().String("ostree-ref", "", `OSTREE reference`)
 	manifestCmd.Flags().String("ostree-parent", "", `OSTREE parent`)
 	manifestCmd.Flags().String("ostree-url", "", `OSTREE url`)
+	manifestCmd.Flags().Bool("use-librepo", true, `use librepo to download packages (disable if you use old versions of osbuild)`)
```
supakeen, line 202: We can 'just' bump the minimum required `osbuild` version instead of this remark; let's do that.

supakeen (approved):
Cool, thank you :)

image-builder-cli.spec
```diff
@@ -3,8 +3,12 @@
 %bcond_with tests
 %bcond_with relax_requires
 
-# The minimum required osbuild version
-%global min_osbuild_version 129
+# The minimum required osbuild version, note that this used to be 129
+# but got bumped to 138 for librepo support which is not strictly
+# required. So if this needs backport to places where there is no
+# recent osbuild available we could simply make --use-librepo false
+# and go back to 129.
```
achilleas-k, line 10: Is this a realistic issue?  I guess it might be if you're `go install`ing ib-cli, but in that case, you're on your own and the spec file doesn't matter.

I mean, there's no way this spec file will be used to **officially** build ib-cli in a distro where a new enough osbuild wont be, so I think this note isn't relevant.

I might be missing something though.

achilleas-k (approved):
The first two commits could be squashed.  The commit message for the second one seems like it's talking about both.

Not a big issue though.

Tiny comment about a comment below.  Overall LGTM.

PR #52, opened 2025-01-08
spec: copy the embedded repo files from the images library

mvo5:
[edit: draft as I'm sure there are issues with the rpm part, someone who knows packit/mockbuild to the rescue! (looking at you @supakeen ;), feel free to force push/fixup! commit here]

Now that we have merged images PR#1112 and have access to the
repository files via the vendor/ directory start with a very
simple spec update and copy the files into our own search
path. Note that once reporegistry is fixed we will not embed
the file in the rpm at all, we will just go:embed.

This also updates the search path for repositories to:
```
var defaultDataDirs = []string{
        "/etc/image-builder",
        "/usr/share/image-builder",
}
```


mvo5:
Its actually less work and less fiddly to just use go:embed and DTRT from the start, sorry for the noise, closing this one.

PR #53, opened 2025-01-08
main: add ostree integration

mvo5:
[draft as this needs tests, input what ostree image could be used for an integration test most welcome]

This commit adds integration for the ostree options. It is modelled loosely after weldr-client/composer-cli and the
```
start-ostree --{ref,parent,url}
```
and uses
```
--ostree-{ref,parent,url}
```


ondrejbudai (commented):
How does this behave if --ostree* is given to a non-ostree image type? We should probably return an error. Same for the opposite case - no ostree options for an ostree-enabled image type.

mvo5:
> How does this behave if --ostree* is given to a non-ostree image type? We should probably return an error. Same for the opposite case - no ostree options for an ostree-enabled image type.

Thanks for the review! I added two tests that ensure the cases are handled correctly.

supakeen:
Seems like there are some actual test case failures while building the RPMs:

```
--- FAIL: TestManifestIntegrationOstreeSmokeErrors (0.06s)
    main_test.go:299: 
        	Error Trace:	/builddir/build/BUILD/image-builder-cli-4/_build/src/github.com/osbuild/image-builder-cli/cmd/image-builder/main_test.go:299
        	Error:      	Error message not equal:
        	            	expected: "iot-raw-image: ostree commit URL required"
        	            	actual  : "cannot find image for: distro:\"fedora-40\" type:\"iot-raw-image\" arch:\"s390x\""
        	Test:       	TestManifestIntegrationOstreeSmokeErrors
```

supakeen (approved):

PR #54, opened 2025-01-08
test: re-enable go/pytest workflow

mvo5:
The go/pytest workflows are curently not run and this PR is here to figure out why :( Last run 11h ago according to https://github.com/osbuild/image-builder-cli/actions/workflows/go.yaml


.pre-commit-config.yaml
```diff
@@ -19,10 +19,10 @@ repos:
     hooks:
       - id: pyspelling
         args: ["--config", ".spellcheck.yml"]
-  - repo: https://github.com/golangci/golangci-lint
-    rev: v1.59.1
-    hooks:
-      - id: golangci-lint
+  # - repo: https://github.com/golangci/golangci-lint
```
mvo5, line 22: I will open a followup that re-enables this but it's a bit annoying, this fails in CI with:
```
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 3
level=warning msg="[runner] Can't run linter goanalysis_metalinter: buildir: failed to load package gpgme: could not load export data: no export data for \"github.com/proglottis/gpgme\""
level=error msg="Running error: can't run linter goanalysis_metalinter\nbuildir: failed to load package gpgme: could not load export data: no export data for \"github.com/proglottis/gpgme\""
```
but not locally and the error gives no indication what is wrong or why it is wrong. Note that gpgme is an indrect dependency only

PR #55, opened 2025-01-09
repos: lookup <buildin>,  /{etc,usr/share}/image-builder/repositories

mvo5:
This commit switches the default repositories to use the build-in
ones from images PR#1112. It is still possible to override using
`/etc/image-builder` and `/usr/share/image-builder` or `--datadir`.

This is implicitly tested via the container test that no longer
includes the files from osbuild-composer in the container.


supakeen (approved):
I like it; perhaps an easier way to provide repositories can be used than only `--datadir` later on.

mvo5:
> I like it; perhaps an easier way to provide repositories can be used than only `--datadir` later on.

Yeah, I'm not attached to --datadir and it was mostly there because of the way reporegistry works. What do you have in mind? just --repositories (so that we can avoid the requirement for the `repository/` prefix)? Or using a yaml repository (and converting it to our own internal json)?

PR #57, opened 2025-01-10
Makefile: fix Packit version by using git tag

schuellerf:
For releases this does not change anything, as the version in the spec file is the same as the git tag. For
intermediate/tainted versions packer seems to patch the spec file to the version given in `.packit.yaml`
(which is the latest git-tag), then the filename and version would be inconsistent.

Rationale for this change was, that the packit build failed, as the version was set "4" but the source-tar-ball version was "5" from the spec file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ GOLANGCI_COMPOSER_IMAGE=composer_golangci
 #         pre-fetched but evaluated at time of use.
 #
 
-VERSION := $(shell (cd "$(SRCDIR)" && grep "^Version:" image-builder-cli.spec | sed -n 's/^[^0-9]*\([1-9][0-9]*\(\.[1-9][0-9]*\)*\)/\1/p'))
+VERSION := $(shell ( git describe --tags --abbrev=0 2>/dev/null || echo v1 ) | sed 's|v||')
 COMMIT = $(shell (cd "$(SRCDIR)" && git rev-parse HEAD))
 PACKAGE_NAME_VERSION = image-builder-cli-$(VERSION)
 PACKAGE_NAME_COMMIT = image-builder-cli-$(COMMIT)


schuellerf:
btw. any pros/cons to switch from
`git describe --tags --abbrev=0`
to
`git describe --tags`
as this `--abbrev=0` was just copied over from https://github.com/osbuild/osbuild/blob/main/.packit.yaml#L18
I guess but I'd really prefer, that source tarballs from non-tagged commits should reflect that in the version and filenames…

Makefile
```diff
@@ -47,7 +47,7 @@ GOLANGCI_COMPOSER_IMAGE=composer_golangci
 #         pre-fetched but evaluated at time of use.
 #
 
-VERSION := $(shell (cd "$(SRCDIR)" && grep "^Version:" image-builder-cli.spec | sed -n 's/^[^0-9]*\([1-9][0-9]*\(\.[1-9][0-9]*\)*\)/\1/p'))
+VERSION := $(shell ( git describe --tags --abbrev=0 2>/dev/null || echo v1 ) | sed 's|v||')
```
thozza, line 50: I don't think that this is the correct fix.

When I look at the osbuild repository, `setup.py` has version `138`, SPEC file has `138` and the latest tag is `v137`.

Makefile uses https://github.com/osbuild/osbuild/blob/fd19ab41fb5a84a15560ca319d2dab4f59c06e27/Makefile#L50

And packit config contains exactly the same way to determine the current version as the one in this repository.

TL;DR; I think we should keep using the Version from the SPEC, which is the next "to be released" version.

thozza: OK, I'm probably wrong, since from looking at osbuild COPR build logs for commits on `main`, I can see that the `package_version` is based on the last tag https://download.copr.fedorainfracloud.org/results/@osbuild/osbuild/fedora-41-x86_64/08476002-osbuild/builder-live.log.gz

The problem is probably that since Packit does not control tarball generation in this case, it can't control the version in its filename.

thozza:
> btw. any pros/cons to switch from `git describe --tags --abbrev=0` to `git describe --tags` as this `--abbrev=0` was just copied over from https://github.com/osbuild/osbuild/blob/main/.packit.yaml#L18 I guess but I'd really prefer, that source tarballs from non-tagged commits should reflect that in the version and filenames…

This may clash with what will be specified in the SPEC file for `Source0:`, since it probably expects the `Version` value in the filename. That's why Packit defines the `%dist` tag as a value containing the git ref and not the version.

thozza (approved):
After going through the Packit COPR logs in osbuild and thinking about this more, I think this is the correct approach.

schuellerf:
Just saw in the official documentation, also just `echo`-ing the name is also valid
https://packit.dev/docs/configuration/examples#custom-archive-creation-for-srpm
…tempted to try this without `--abbrev=0` 😏

mvo5:
Thanks! Lets fix as is an we can always do a followup, unblocking rpms is super nice!

thozza:
> Just saw in the official documentation, just `echo`-ing the name is also valid https://packit.dev/docs/configuration/examples#custom-archive-creation-for-srpm …tempted to try this without `--abbrev=0` 😏

Go ahead :wink: 

AFAIK, this is an `rpmbuild` thing and not Packit specific. Since `rpmbuild` will look for a specific file in the `SOURCES` directory, based on what `Source0: %{gosource}` expands to.

schuellerf:
> Go ahead 😉
> 
> AFAIK, this is an `rpmbuild` thing and not Packit specific. Since `rpmbuild` will look for a specific file in the `SOURCES` directory, based on what `Source0: %{gosource}` expands to.

@thozza my question would be - do you expect problems if the version is `4-12-g26cb415` (dashes, letter "g" …) instead of `4` (which IMHO is plain _wrong_ if it's not exactly the tag `v4`)

PR #58, opened 2025-01-10
Makefile/packit: Set version to "single number" only for releases

schuellerf:
RPMs and releases should reflect if it's built from a release tag or some other commit.

Expected behavior is, that if you build a tag like "v4" the result is `image-builder-cli-4.tar.gz` otherwise it should visualize that this is NOT a release e.g. `image-builder-cli-4.12.g30166b0.tar.gz`

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,7 +16,7 @@ srpm_build_deps:
 
 actions:
   # v1 just the default prior to having tags
-  get-current-version: bash -c "( git describe --tags --abbrev=0 2>/dev/null || echo v1 ) | sed 's|v||'"
+  get-current-version: bash -c "make show-version"
   post-upstream-clone: bash -c "go mod vendor && ./tools/rpm_spec_add_provides_bundle.sh"
   create-archive: bash -c "make release_artifacts"
 
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ GOLANGCI_COMPOSER_IMAGE=composer_golangci
 #         pre-fetched but evaluated at time of use.
 #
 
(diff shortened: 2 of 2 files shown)


schuellerf:
Seems to work - I'd really prefer this, if there is no drawback…
see

https://download.copr.fedorainfracloud.org/results/packit/osbuild-image-builder-cli-58/fedora-41-x86_64/08498668-image-builder-cli/

mvo5:
Do we have the version now twice? Looking at e.g. https://dashboard.packit.dev/jobs/copr/2118828 I see:
```
sudo dnf install -y image-builder-cli-4.13.gf402972-1.20250110124024754480.pr58.13.gf402972.el10.aarch64
```
or am I missing something?

schuellerf:
I think the filename might be constructed from "version" and "release"
```
Version     : 4.13.gf402972
Release     : 1.20250110124024754480.pr58.13.gf402972.fc40
```

schuellerf:
For me the key would be that "upgrading" and "downgrading" would work better? Not sure if "release" is considered here by rpm/dnf.

schuellerf:
…but you might be right - if "release" is respected properly - this PR is rendered an overkill :-/

supakeen:
Release is considered and this is (almost) exactly what it is meant for. We're building version 4; with some additional commits on top, see it as if we're applying patches to version 4.

thozza:
RPM filenames are **always** constructed as NVRA - name-version-release.arch. The full NVRA is taken into consideration for upgrades and downgrades.

Personally, I use COPR builds from `main` for our tools on my system and don't like the NVRA that's used as a result of this PR. 

It is common in RPM world to use the `Release` to distinguish changes made to the latest upstream release. In this context, using a single digit version and adding the git ref to the Release is IMHO the common approach.

PR #63, opened 2025-01-13
README: add blueprint example

mvo5:
This commit adds an example for how to pass the blueprint and what content the blueprint should have.


ondrejbudai (approved):
Thanks!

Issue #31, opened 2024-12-20
Move `build` from positional args to named.


supakeen:
It'd be nicer if the `image-builder build` subcommand takes named arguments instead of positional arguments. Currently it looks like this:

```
image-builder build --distro fedora-41 minimal-raw blueprint.toml
```

I propose:

```
image-builder build --distro fedora-41 --type minimal-raw --blueprint blueprint.toml
```

This would allow for multi-type builds (a very strong advantage of the `osbuild` system):

```
image-builder build --distro fedora-41 --type iso --type qcow2 --blueprint blueprint.toml
```

And would allow us to in the future have other inputs except the built-in image types such as a custom YAML file for custom images.

mvo5:
Fwiw, I like that it's more "uniform" this way, i.e. to me it feels a bit unbalanced that we currently have "--distro centos-9" but type is just an positional argument and has no --type. But then its a bit more typing for our users. 

We should probably have a brainstorm again in the new year.

achilleas-k:
I agree, there's a convenience to positional args, but I think once you go beyond 3, it starts becoming easier to forget which is which.  I think the extra typing is an acceptable tradeoff for explicitness.  I wouldn't be against going that way.

mvo5:
@ondrejbudai Any (strong) opinions here?

ondrejbudai:
I only have one strong opinion: the user experience for the first time users must be as simple as possible. I like that positional arguments is all you need to get an image, I find it extremely user-friendly: `image-builder build qcow2`.

Requiring a named argument makes the command longer, and imo slightly harder to document.

---

Would it make sense to split these use-cases into 2 subcommands? One very simple one (`build`) for the masses, one complex one for people that need the complexity to succeed. I'm even thinking that the complex command shouldn't accept arguments, but rather a file (might be an overkill?):

```console
$ cat fedora-iot.yaml
ostree-url: example.com/fedora
ostree-ref: fedora/rawhide
types:
  - iot-installer
  - iot-simplified-installer
  - iot-raw
  - iot-container
distro: fedora-rawhide
$ image-builder-cli multi-build fedora-iot.yaml
```

supakeen:
I'm fine with either approach; I created a quick PR to have `--type` because it was very little work to have it.

I'll add an opinion that if we do what @ondrejbudai proposes to please call these commands `build-multi` (so the reverse) so they sort correctly (all under `build` :)).

supakeen:
[...]

PR #65, opened 2025-01-14
`--blueprint` (HMS-5340)

supakeen:
Changes the blueprint argument to a named argument after some discussion outside of GitHub. This allows for passing multiple image types later on and slightly simplifies the case of not passing a blueprint.

See #31.

/jira-epic COMPOSER-2418

JIRA: [HMS-5340](https://issues.redhat.com/browse/HMS-5340)

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -84,15 +84,18 @@ func cmdManifestWrapper(cmd *cobra.Command, args []string, w io.Writer, archChec
 		return nil, err
 	}
 
-	var blueprintPath string
-	imgTypeStr := args[0]
-	if len(args) > 1 {
-		blueprintPath = args[1]
+	blueprintPath, err := cmd.Flags().GetString("blueprint")
+	if err != nil {
+		return nil, err
 	}
+
+	imgTypeStr := args[0]
+
 	bp, err := blueprintload.Load(blueprintPath)
 	if err != nil {
 		return nil, err
 	}
+
 	distroStr, err = findDistro(distroStr, bp.Distro)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


ondrejbudai (changes requested):
The readme example with a blueprint needs an update.

cmd/image-builder/main.go
```diff
@@ -84,15 +84,18 @@ func cmdManifestWrapper(cmd *cobra.Command, args []string, w io.Writer, archChec
 		return nil, err
 	}
 
-	var blueprintPath string
 	imgTypeStr := args[0]
-	if len(args) > 1 {
-		blueprintPath = args[1]
+
+	blueprintPath, err := cmd.Flags().GetString("blueprint")
```
mvo5, line 89: (super nitpick) having all Flags().Get() group together feels slightly nicer, so moving imgTypeStr := args[0] after the blueprint get, i.e.:
```go
diff --git a/cmd/image-builder/main.go b/cmd/image-builder/main.go
index adc44a0..834ec30 100644
@@ -83,13 +83,11 @@ func cmdManifestWrapper(cmd *cobra.Command, args []string, w io.Writer, archChec
        if err != nil {
                return nil, err
        }
-
-       imgTypeStr := args[0]
-
        blueprintPath, err := cmd.Flags().GetString("blueprint")
        if err != nil {
                return nil, err
        }
+       imgTypeStr := args[0]
 
        bp, err := blueprintload.Load(blueprintPath)
        if err != nil {

```
but that is a super minor issue and I can do that at some later point.

mvo5 (dismissed):
Thank you, one super nitpick inline but not a blocker so feel free to merge.

cmd/image-builder/main.go
supakeen: I've moved it directly; no need for a followup.

supakeen:
> The readme example with a blueprint needs an update.

Good catch, done.

achilleas-k (approved):
TY.  LGTM!

ondrejbudai (approved):
🚢 it!

PR #68, opened 2025-01-15
pre-commit: re-enable golangci-lint

schuellerf:
Followup of #54

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -12,8 +12,12 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
+      - name: Update apt repositories
+        run: sudo apt update
       - name: Install lint dependencies
         run: sudo apt install -y yamllint aspell pre-commit
+      - name: Install project dependencies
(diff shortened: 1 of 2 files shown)


schuellerf:
@mvo5 seems to work when adding the libraries…
I'm not sure if those are installed with the `sudo dnf install osbuild osbuild-depsolve-dnf` mentioned in the `README.md` or we should add the libraries there, too?

achilleas-k:
> I'm not sure if those are installed with the `sudo dnf install osbuild osbuild-depsolve-dnf` mentioned in the `README.md` or we should add the libraries there, too?

They're not.  gpgme isn't a dependency of osbuild itself, but of images (again, indirectly):
```
› go mod why github.com/proglottis/gpgme
# github.com/proglottis/gpgme
github.com/osbuild/image-builder-cli/internal/manifestgen
github.com/osbuild/images/pkg/container
github.com/containers/image/v5/signature
github.com/proglottis/gpgme
```

I don't know what exactly golangci-lint does with imports, but the FAQ does say:
>  If using CGO, ensure all require system libraries are installed.
which gpgme does.

mvo5 (approved):
Thank you

Issue #46, opened 2025-01-06
SBOM exports


supakeen:
Let's allow exporting SBOMs together with the build pipelines. They can end up in the same output directory next to the produced artifact.

Probably related to #45, #43, or/and #42.

Can be optional, e.g. `image-builder build --with-sbom ...`.

PR #69, opened 2025-01-16
ibcli: add new --extra-artifacts option that can write an sbom file

mvo5:
This commit adds an option to generate an sbom file via the
new `--export-sbom`option.

This to follows https://github.com/osbuild/osbuild-composer/pull/4359 and names the isboms `<image_filename>.<pipeline_purpose>-<pipeline_name>.spdx.json`

Once that is in we can do `--extra-artifacts=manifest` (this should be much simpler) and maybe later `extra-artifacts=buildlog`.

Closes: https://github.com/osbuild/image-builder-cli/issues/46

JIRA: RHELBU-3011

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -14,4 +14,7 @@ qcow
 UEFI
 datadir
 copr
+SBOM
+SBOMs
+spdx
 toml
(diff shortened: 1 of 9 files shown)


supakeen (approved):
This is a really good change and a great start to start iterating on. There's some minor things I'd do differently or change but nothing that blocks having this cool feature :)

PR #71, opened 2025-01-17
spec: require osbuild

supakeen:
We can't build images without having osbuild installed.

### DIFF
--- a/image-builder-cli.spec
+++ b/image-builder-cli.spec
@@ -52,6 +52,7 @@ BuildRequires:  btrfs-progs-devel
 %endif
 
 Requires:       osbuild
+Requires:       osbuild-depsolve-dnf
 
 %description
 %{common_description}
--- a/image-builder-cli.spec
+++ b/image-builder-cli.spec
@@ -51,11 +51,11 @@ BuildRequires:  btrfs-progs-devel
 # BUNDLE_END
 %endif
 
-Requires:   osbuild
-Requires:   osbuild-ostree
-Requires:   osbuild-lvm2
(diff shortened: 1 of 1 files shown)


image-builder-cli.spec
```diff
@@ -51,6 +51,8 @@ BuildRequires:  btrfs-progs-devel
 # BUNDLE_END
 %endif
 
+Requires:       osbuild
```
mvo5, line 54: Do we also need osbuild-depsolve-dnf ?

mvo5 (commented):
Thank you! One quick question inline before I approve

image-builder-cli.spec
supakeen: You're too quick :D

supakeen: Ok, now everything is listed including the minimal osbuild version we already had.

mvo5 (approved):
Thank you! Looks great, one wondering about packaging in the rpm world but lets not block on it as we can always tweak it (its just in copr right now)

image-builder-cli.spec
```diff
@@ -51,6 +51,12 @@ BuildRequires:  btrfs-progs-devel
 # BUNDLE_END
 %endif
 
+Requires:   osbuild >= %{min_osbuild_version}
+Requires:   osbuild-ostree >= %{min_osbuild_version}
```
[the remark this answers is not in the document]
supakeen: I took them from the requires in osbuild-composer; assuming we want to be 'kinda the same' at least initially :)

PR #72, opened 2025-01-17
fix up usage of cmd

supakeen:
Fixes up the usage strings and argument counts for the build and manifest subcommands after we moved to `--blueprint`.


mvo5 (approved):
Nice catch

Issue #42, opened 2025-01-06
Store manifest together with artifact


supakeen:
When building an image with:

```
sudo podman run --privileged \           
   -v ./output:/output \
   ghcr.io/osbuild/image-builder-cli:latest \
   build \
   --distro fedora-41 \
   minimal-raw
```

The output directory only contains the artifact produced by the `minimal-raw` image type.

I'd like it to also (optionally?) contain the manifest that was used for the build so I can later rebuild the same image or inspect the image that was built.

PR #80, opened 2025-01-20
main: add `--extra-artifacts=manifest`

mvo5:
This commit adds support for `--extra-artifacts=manifest`. If
that is given as part of the build an extra artifacts called
`.osbuild-manifest.json` will be created in the
output directory.

Closes: https://github.com/osbuild/image-builder-cli/issues/42


supakeen (approved):
Very cool, thank you. Perhaps we could mention SBOMs in the README as well (now that we mention `manifest`) but that can be done in a followup.

PR #87, opened 2025-01-22
cmd/image-builder/main: improve wording of `--blueprint` help

schuellerf:
When reading `--blueprint string` I briefly thought it might also be possible to put a whole blueprint content as string into image-builder-cli argument 😕

This wording should avoid misunderstandings.


supakeen:
I'd like to bikeshed a tiny bit: filename or path? 🏚️

schuellerf:
The one which can be understood better by users 🤷‍♂️

mvo5 (approved):
Thank you, nice catch - this makes sense to me! (no opinion on filename vs path)

PR #88, opened 2025-01-22
specfile: fedora review changes

supakeen:
Changes as pointed out in the [Fedora package review](https://bugzilla.redhat.com/show_bug.cgi?id=2339108).

### DIFF
--- a/image-builder.spec
+++ b/image-builder.spec
@@ -19,7 +19,7 @@ OSTree commits. Uses osbuild under the hood.
 Name:           image-builder
 Release:        1%{?dist}
 Summary:        An image building executable using osbuild
-ExcludeArch:    i686 armv7hl
+ExcludeArch:    i686
 
 # Upstream license specification: Apache-2.0
 # Others generated with:
--- a/image-builder.spec
+++ b/image-builder.spec
@@ -111,5 +111,5 @@ cd $PWD/_build/src/%{goipath}
 # the changelog is distribution-specific, therefore there's just one entry
 # to make rpmlint happy.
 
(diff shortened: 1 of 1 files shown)


ondrejbudai (changes requested):
For the record, the only concerning thing about this PR is that the `Source0` is now hardcoded to bundled releases (as required by the Fedora Packaging Guideline).

We were wondering about how packit handles it in its COPR integration. It seems like it doesn't actually care about the original `Source0`, and it just overrides it with whatever is needed to build a branch/PR. I verified this by downloading the SRPM, and inspecting the embedded tarball.

Just one thing: @supakeen, can you try removing the version change? In all other projects, we have LATEST_VERSION+1 in the specfile on main, so it would be great to do this here as well (otherwise we would have to adjust the automation I think).

supakeen:
> For the record, the only concerning thing about this PR is that the `Source0` is now hardcoded to bundled releases (as required by the Fedora Packaging Guideline).
> 
> We were wondering about how packit handles it in its COPR integration. It seems like it doesn't actually care about the original `Source0`, and it just overrides it with whatever is needed to build a branch/PR. I verified this by downloading the SRPM, and inspecting the embedded tarball.
> 
> Just one thing: @supakeen, can you try removing the version change? In all other projects, we have LATEST_VERSION+1 in the specfile on main, so it would be great to do this here as well (otherwise we would have to adjust the automation I think).

Of course; I've rebased and changed the `Source0` commit to drop the version change which now puts it on the same version as `main` again.

ondrejbudai (approved):
Thanks!

PR #92, opened 2025-01-24
image-builder: drop internal manifestgen in favor of images

mvo5:
This commit drops the internal `manifestgen` package in favor
of using the version in `images` now that
https://github.com/osbuild/images/pull/1153 is merged.


cmd/image-builder/manifest.go
```diff
@@ -53,7 +54,7 @@ func generateManifest(dataDir string, img *imagefilter.Result, output io.Writer,
 		if err := os.MkdirAll(outputDir, 0755); err != nil {
 			return err
 		}
-		manifestGenOpts.SBOMWriter = func(filename string, content io.Reader) error {
+		manifestGenOpts.SBOMWriter = func(filename string, content io.Reader, docType sbom.StandardType) error {
```
supakeen, line 57: Perhaps we want this to be `StandardTypeSpdx` (or instead use `manifestgen.defaultDepsolverSBOMType`?

This might also be the cause of the test failure(s).

supakeen (changes requested):
Tests currently broken, added a suggestion that might fix it.

cmd/image-builder/manifest.go
mvo5: Thanks for raising this! I suspect the test needs an update, I will work on this.

mvo5: This is fixed now, the test needed an update because the original manifestgen code was creating an sbom that contained one level of nesting ("Document") too much (Thozza caught it in the review for images).

supakeen (approved):

PR #93, opened 2025-01-24
github: run tests inside a fedora container

mvo5:
Running the tests inside fedora means we get much more meaningful results as we will have all the required packaes and there will be no need to skip tests.


supakeen (approved):
Looks good to me. Not the biggest fan of the cURL step but running more tests is nice.

Containerbuilds are still failing due to GCC segfaults.

PR #97, opened 2025-01-25
image-builder: create directory

### LINKED ISSUE #94: `build --with-manifest` fails due to not creating the output directory
When running:

```
user@muja image-builder-cli € time sudo go run ./cmd/image-builder build --with-manifest minimal-raw
No distro name specified, selecting "fedora-41" based on host, use --distro to override
WARN[0000] Failed to load consumer certs: no consumer key found 
error: open fedora-41-minimal-raw-x86_64/fedora-41-minimal-raw-x86_64.osbuild-manifest.json: no such file or directory
exit status 1
sudo go run ./cmd/image-builder build --with-manifest minimal-raw  0.01s user 0.01s system 0% cpu 7.861 total
```

The build seems to fails due to the output directory not being created before `image-builder` tries to write the manifest out. Manually creating the `fedora-41-minimal-raw-x86_64` directory allows the build to succeed.

supakeen:
Create the output directory; this is a no-op if it already exists but fixes #94 if it doesn't. The code is similar to (but not identical to) the checks done in `cmd/image-builder/manifest.go` for the `withSBOM` parts.

We might (or I) might want to consolidate those two codepaths either here in a followup.

### DIFF
--- a/cmd/image-builder/build.go
+++ b/cmd/image-builder/build.go
@@ -28,6 +28,9 @@ func buildImage(res *imagefilter.Result, osbuildManifest []byte, opts *buildOpti
 	}
 	if opts.WriteManifest {
 		p := filepath.Join(opts.OutputDir, fmt.Sprintf("%s.osbuild-manifest.json", outputNameFor(res)))
+		if err := os.MkdirAll(filepath.Dir(p), 0755); err != nil {
+			return err
+		}
 		if err := os.WriteFile(p, osbuildManifest, 0644); err != nil {
 			return err
 		}
--- a/test/test_container.py
+++ b/test/test_container.py
@@ -52,3 +52,23 @@ def test_container_manifest_generates_sbom(tmp_path, build_container):
     sbom_json = json.loads(image_sbom_json_path.read_text())
     # smoke test that we have glibc in the json doc
     assert "glibc" in [s["name"] for s in sbom_json["packages"]], f"missing glibc in {sbom_json}"
+
+
+@pytest.mark.skipif(os.getuid() != 0, reason="needs root")
+def test_container_build_generates_manifest(tmp_path, build_container):
+    output_dir = tmp_path / "output"
+    output_dir.mkdir()
+    subprocess.check_call([
+        "podman", "run",
+        "--privileged",
+        "-v", f"{output_dir}:/output",
+        build_container,
+        "build",
+        "minimal-raw",
+        "--distro", "centos-9",
+        "--with-manifest",
+    ], stdout=subprocess.DEVNULL)
+    arch = platform.machine()
+    fn = f"centos-9-minimal-raw-{arch}/centos-9-minimal-raw-{arch}.osbuild-manifest.json"
+    image_manifest_path = output_dir / fn
+    assert image_manifest_path.exists()
--- a/cmd/image-builder/manifest.go
+++ b/cmd/image-builder/manifest.go
@@ -25,6 +25,9 @@ type manifestOptions struct {
 
 func sbomWriter(outputDir, filename string, content io.Reader) error {
 	p := filepath.Join(outputDir, filename)
+	if err := os.MkdirAll(filepath.Dir(p), 0755); err != nil {
+		return err
+	}
 	f, err := os.Create(p)
 	if err != nil {
 		return err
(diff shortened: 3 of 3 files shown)


cmd/image-builder/build.go
```diff
@@ -26,9 +26,21 @@ func buildImage(res *imagefilter.Result, osbuildManifest []byte, opts *buildOpti
 	if opts.OutputDir == "" {
 		opts.OutputDir = outputNameFor(res)
 	}
+
 	if opts.WriteManifest {
-		p := filepath.Join(opts.OutputDir, fmt.Sprintf("%s.osbuild-manifest.json", outputNameFor(res)))
-		if err := os.WriteFile(p, osbuildManifest, 0644); err != nil {
```
mvo5, line 31: Thanks for looking into this and the patch!

Would this:
```go
diff --git a/cmd/image-builder/build.go b/cmd/image-builder/build.go
index 26b6151..b00c9c6 100644
--- a/cmd/image-builder/build.go
+++ b/cmd/image-builder/build.go
@@ -28,6 +28,9 @@ func buildImage(res *imagefilter.Result, osbuildManifest []byte, opts *buildOpti
        }
        if opts.WriteManifest {
                p := filepath.Join(opts.OutputDir, fmt.Sprintf("%s.osbuild-manifest.json", outputNameFor(res)))
+               if err := os.MkdirAll(filepath.Dir(p), 0755); err != nil {
+                       return err
+               }
                if err := os.WriteFile(p, osbuildManifest, 0644); err != nil {
                        return err
                }
```
also fix the issue? Having a small test (or modifiying one of the existing ones) to trigger this issue would be great.

mvo5: Thanks, I have no strong opinion, the two are slightly different but maybe:
```go
diff --git a/cmd/image-builder/manifest.go b/cmd/image-builder/manifest.go
index 7a4ab0c..ac60674 100644
--- a/cmd/image-builder/manifest.go
+++ b/cmd/image-builder/manifest.go
@@ -24,6 +24,9 @@ type manifestOptions struct {
 
 func sbomWriter(outputDir, filename string, content io.Reader) error {
        p := filepath.Join(outputDir, filename)
+       if err := os.MkdirAll(filepath.Dir(p), 0755); err != nil {
+               return err
+       }
        f, err := os.Create(p)
        if err != nil {
                return err
@@ -50,9 +53,6 @@ func generateManifest(dataDir string, img *imagefilter.Result, output io.Writer,
                if outputDir == "" {
                        outputDir = outputNameFor(img)
                }
-               if err := os.MkdirAll(outputDir, 0755); err != nil {
-                       return err
-               }
                manifestGenOpts.SBOMWriter = func(filename string, content io.Reader) error {
                        return sbomWriter(outputDir, filename, content)
                }

```
makes sense, would make the sbomWriter easier to reuse.

supakeen:
@mvo5 I've addressed the testcase, I moved some commits around so the testcase is first (so it fails *before* the rest of the commits are applied). The testcase is in the integration tests.

Aside from that I've evened out the approaches taken in both `sbomWriter` and `buildImage`.

test/test_container.py
```diff
@@ -52,3 +52,23 @@ def test_container_manifest_generates_sbom(tmp_path, build_container):
...
+@pytest.mark.skipif(os.getuid() != 0, reason="needs root")
+def test_container_build_generates_manifest(tmp_path, build_container):
+    output_dir = tmp_path / "output"
+    output_dir.mkdir()
+    subprocess.check_call([
+        "podman", "run",
+        "--privileged",
+        "-v", f"{output_dir}:/output",
+        build_container,
+        "build",
```
mvo5, line 66: Ideally we would use the "build_fake_container" here that mocks osbuild (c.f. https://github.com/osbuild/image-builder-cli/pull/21/commits/13279a2562eb3b4b3ea0932c4ca0c322f3baa12a) as for this particular test we don't really need the full build. But its fine, we can tweak it after the other one is merged (or if you prefer you could cherry pick that commit and replace "build_container" with "build_fake_container" in the test and it should run a lot quicker :)

mvo5 (approved):
Thank you, this looks very nice now and extra thanks for the test. One tiny suggestion inline but that can be a followup

test/test_container.py
supakeen: I think we do; because if we don't then the actual error (that was there) isn't triggered? Perhaps if we mock the filesystem as well; but that'd be a followup.

Issue #98, opened 2025-01-25
Consistency in named arguments


supakeen:
We have a few arguments that all deal with filesystem paths:

1. `--blueprint`
2. `--datadir`
3. `--output-dir`

These either have `dir`, `-dir`, or nothing special in the name. I'd like to make them predictable and my suggestion is to drop the `dir` and `-dir` parts of the names. So we get:

1. `--blueprint`
2. `--data`
3. `--output`

Thoughts? `--data` might be a bit terse and in general is pretty non-descriptive; perhaps `--defs` suits better to align with the name that seems to be the one we want in `images`?

`--output` clashes with the `--output` on `list-images`, I'd suggest renaming that one to `--format`.

A full command line would then look like (for example):

```
image-builder build --defs=~/my-defs --blueprint=foo.toml --output=/opt/build qcow2
```

supakeen:
Demo implementation in #99 so it can be tried out to see how it feels; each change as a separate commit so they can be cherry-picked for those we like.

supakeen:
Implementation has landed in #138.

PR #99, opened 2025-01-25
argument naming consistency

supakeen:
This implements my ideas from #98. Each change is a separate commit so they can be dropped or modified.

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -18,3 +18,4 @@ SBOM
 SBOMs
 spdx
 toml
+defs
--- a/README.md
+++ b/README.md
@@ -183,9 +183,9 @@ A: The osbuild binary is used to actually build the images but beyond that
 
 Q: Can I have custom repository files?
 A: Sure! The repositories are encoded in json in "<distro>-<vesion>.json",
-   files, e.g. "fedora-41.json". See these [examples](https://github.com/osbuild/images/tree/main/data/repositories). Use the "--datadir" switch and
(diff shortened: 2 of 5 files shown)


mvo5:
Thanks for the PR. This is a tricky one. I agree with the concerns. Its also a bit of a can of worms:

1. The `--datadir` is a bit annyoing, the only thing we currently can customize here is the repositories and its only called like this because the `reporegistry.New()` will always append `repositories/`to the dirs passed into reporegistry.New(). So we could  fix reporegistry.New() to take (the direct repo config paths) and then we could rename this to `--repositories`. But in the future with the image defintion in yaml this will most likely point to those as well so probably worth keeping. I not super convinved by --defs though, its nice and short but lacks the hint of that its a "rootdir" for a bunch of data.
2. The `--output` feels a bit too generic. It could mean, file, dir or format (like we use it now). As we generally only support specifying a dir having "dir" in it seems okay. We will (hopefully) also get supprot to change the output-filenamename of the imagefile itself (instead of the current hardcoding) and then this might also become a bit confusing.
3. Using "format" instead of "output" sounds nice, I think its early enough to do it and it seems cleaner to me (and its early enough to change this still I feel)
4. blueprint vs blueprint-path is a tricky one too, the later is better in the "more-precise" sense, the former is shorter and the help explain it but maybe not enough? I slightly lean towards shorter.

Sorry, many thoughts, naming is hard.

achilleas-k:
Not sure about `--defs`.  We use it a lot internally, but I have no idea how (image/distro) "definitions" sounds to a user, if it sounds like what it is.

supakeen:
Since we had a discussion about this I'll re-open a new PR with the discussed names instead of updating this one.

PR #108, opened 2025-01-29
Containerfile: build without CGO to avoid arm64 crashing

mvo5:
This commit changes the build to not build with CGO. There is really no need to build with CGO, we only need it for the `crypt()` version in the images library. However since with the merge of https://github.com/osbuild/images/pull/1114
that is no longer an issue.


achilleas-k (approved):
LGTM

supakeen (approved):

PR #111, opened 2025-01-30
main: silence all logrus logging for now

mvo5:
This commit silences logrus logging for now by just dicarding everything from it. The rational is that the underlying libraries produce a lot of output, e.g. every build currently results in
```console
$ image-builder build qcow2
WARN[0000] Failed to load consumer certs: no consumer key found
...
```
which is confusing to our users. What is worse is that containers also uses logrus so we can get random warnings without context from there too. Until we have a way to filter log messages this seems to be the most practical way.

We can add `--verbose` or `--debug` later to enable these kinds of messages.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -203,9 +203,13 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 }
 
 func run() error {
-	// images logs a bunch of stuff to Debug/Info that is distracting
-	// the user (at least by default, like what repos being loaded)
-	logrus.SetLevel(logrus.WarnLevel)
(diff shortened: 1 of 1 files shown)


supakeen (approved):
Agreed; I'll open a ticket to track a `--verbose` (or multiple `-v`'s).

lzap:
This is fine, I plan to refactor composer/images to make use of the stdlib `log/slog` but that is the same story and a similar solution.

Alternatively you can grab my [ring buffer writer implementation from a different project](https://github.com/lzap/packer-plugin-image-builder/blob/main/cmd/plugin/tail_writer.go) and use it here storing last N lines, when error occurs you can pickup the data from the ring buffer and display it.

PR #116, opened 2025-02-03
ci/packit: set downstream name

supakeen:
Since our upstream repository name differs from the downstream name we should set the `downstream_package_name` [1] config option.

[1]: https://packit.dev/docs/configuration#downstream_package_name


ondrejbudai (approved):
Thanks!

PR #117, opened 2025-02-03
specfile: build requires `libxcrypt-compat`

supakeen:
We received a commit downstream by Björn Esser. After asking around for the reasoning behind the commit it seems that we're likely assuming that this package is available in the buildroot.

Let's depend on it explicitly instead of assuming it is there.


ondrejbudai (approved):
Thanks!

PR #119, opened 2025-02-03
main: add new upload command

mvo5:
[edit: one open question is if this PR should also contain a variant of this like `build --uplload-to` so that build/upload can be done in a single call.

This commit adds a new `upload` command that can be used to upload a raw image to the cloud. Currently only AWS is supported but as images adds more clouds to the uploader interfac we can trivially expand more.

Note that this needs https://github.com/osbuild/images/pull/1185

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,22 @@ It is possible to generate spdx based SBOM (software bill of materials)
 documents as part of the build. Just pass `--with-sbom` and
 it will put them into the output directory.
 
+### Cloud integration
+
+When building an image type that can be uploaded to the cloud
+(e.g. an "ami") image-builder will automatically upload if
+all cloud parameters are provided, e.g.
+```
+$ image-builder build ami --distro centos-9 \
+    --aws-region us-east-1 \
+	--aws-bucket example-bucket \
+	--aws-ami-name my-image-1
+```
+Images can also be uploaded with the `image-builder upload` command
+after they are built.
+
+
+
 ### Filtering
 
 When listing images, it is possible to filter:
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -9,13 +10,11 @@ import (
 	"path/filepath"
 	"syscall"
 
-	"github.com/cheggaaa/pb/v3"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/osbuild/bootc-image-builder/bib/pkg/progress"
 	"github.com/osbuild/images/pkg/arch"
-	"github.com/osbuild/images/pkg/cloud/awscloud"
 	"github.com/osbuild/images/pkg/imagefilter"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/ostree"
@@ -176,66 +175,6 @@ func progressFromCmd(cmd *cobra.Command) (progress.ProgressBar, error) {
 	return progress.New(progressType)
 }
 
-var awscloudNewUploader = awscloud.NewUploader
-
-func cmdUploadAWS(cmd *cobra.Command, args []string) error {
-	amiName, err := cmd.Flags().GetString("aws-ami-name")
-	if err != nil {
-		return err
-	}
-	bucketName, err := cmd.Flags().GetString("aws-bucket")
-	if err != nil {
-		return err
-	}
-	region, err := cmd.Flags().GetString("aws-region")
-	if err != nil {
-		return err
-	}
-
-	rawDiskPath := args[0]
-	// XXX: can we actually inspect the image or leave some artifacts?
-	if filepath.Ext(rawDiskPath) != ".raw" {
-		return fmt.Errorf("expecting a raw disk ending with '.raw', got %q", filepath.Base(rawDiskPath))
-	}
-
-	uploader, err := awscloudNewUploader(region, bucketName, amiName, nil)
-	if err != nil {
-		return err
-	}
-	f, err := os.Open(rawDiskPath)
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-
-	// setup basic progress
-	st, err := f.Stat()
-	if err != nil {
-		return fmt.Errorf("cannot stat upload: %v", err)
-	}
-	pbar := pb.New64(st.Size())
-	pbar.Set(pb.Bytes, true)
-	pbar.SetWriter(osStdout)
-	r := pbar.NewProxyReader(f)
-	pbar.Start()
-	defer pbar.Finish()
-
-	return uploader.UploadAndRegister(r, osStderr)
-}
-
-func cmdUpload(cmd *cobra.Command, args []string) error {
-	uploadTo, err := cmd.Flags().GetString("to")
-	if err != nil {
-		return err
-	}
-	switch uploadTo {
-	case "aws":
-		return cmdUploadAWS(cmd, args)
-	default:
-		return fmt.Errorf("unsupported cloud %q", uploadTo)
-	}
-}
-
 func cmdBuild(cmd *cobra.Command, args []string) error {
 	cacheDir, err := cmd.Flags().GetString("cache")
 	if err != nil {
@@ -274,13 +213,49 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var uploadUnsupported *UploadTypeUnsupportedError
+	var missingUploadConfig *MissingUploadConfigError
+	uploader, err := uploaderFor(cmd, res.ImgType.Name())
+	if err != nil && !errors.As(err, &missingUploadConfig) && !errors.As(err, &uploadUnsupported) {
+		return err
+	}
+	if missingUploadConfig != nil && !missingUploadConfig.allMissing {
+		return fmt.Errorf("partial upload config provided: %w", err)
+	}
+
+	if uploader != nil {
+		pbar.SetPulseMsgf("Checking cloud access")
+		if err := uploaderCheckWithProgress(pbar, uploader); err != nil {
+			return err
+		}
+	}
+
+	// XXX: support output filename via commandline (c.f.
+	//   https://github.com/osbuild/images/pull/1039)
+	if outputDir == "" {
+		outputDir = outputNameFor(res)
+	}
 	buildOpts := &buildOptions{
 		OutputDir:     outputDir,
 		StoreDir:      cacheDir,
 		WriteManifest: withManifest,
 	}
 	pbar.SetPulseMsgf("Image building step")
-	return buildImage(pbar, res, mf.Bytes(), buildOpts)
+	if err := buildImage(pbar, res, mf.Bytes(), buildOpts); err != nil {
+		return err
+	}
+
+	if uploader != nil {
+		// XXX: integrate better into the progress, see bib
+		pbar.Stop()
+		imagePath := filepath.Join(outputDir, res.ImgType.Name(), res.ImgType.Filename())
+
+		if err := uploadImageWithProgress(uploader, imagePath); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func cmdDescribeImg(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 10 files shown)


cmd/image-builder/main.go
```diff
@@ -156,6 +159,63 @@ func cmdManifest(cmd *cobra.Command, args []string) error {
...
+	region, err := cmd.Flags().GetString("aws-region")
+	if err != nil {
+		return err
+	}
+
+	rawDiskPath := args[0]
+	// XXX: can we actually inspect the image or leave some artifacts?
+	if filepath.Ext(rawDiskPath) != ".raw" {
+		return fmt.Errorf("expecting a raw disk ending with '.raw', got %q", filepath.Base(rawDiskPath))
+	}
```
achilleas-k: Not sure about this.  Yeah, `.raw` is conventional but do we really want to enforce it?  Especially since we have plans to let people set the name for the artifact, and I've seen cases where people will choose `.img` or similar.  It'd be much nicer to look for the PT header if we want to do it right.  Either that or not check it at all.

README.md
```diff
@@ -114,6 +114,22 @@ It is possible to generate spdx based SBOM (software bill of materials)
 documents as part of the build. Just pass `--with-sbom` and
 it will put them into the output directory.
 
+### Cloud integration
+
+When building an image type that can be uploaded to the cloud
+(e.g. an "ami") image-builder will automatically upload if
+all cloud paramters are provided, e.g.
```
achilleas-k, line 121: Typo: parameters

README.md
```diff
@@ -114,6 +114,22 @@ It is possible to generate spdx based SBOM (software bill of materials)
 documents as part of the build. Just pass `--with-sbom` and
 it will put them into the output directory.
 
+### Cloud integration
+
+When building an image type that can be uploaded to the cloud
+(e.g. an "ami") image-builder will automatically upload if
```
achilleas-k, line 120: "that can be uploaded to the cloud" isn't a bad way of describing it but I'd rather be more technically accurate in this kind of documentation.  Something like:
> When building an image that's meant to work with a cloud platform ...

or

> When building an image that targets a cloud provider/platform ...

Something like that?

achilleas-k (changes requested):
No major deal breakers, but requesting changes for typo and to discuss the `.raw` extension requirement.

cmd/image-builder/main.go
achilleas-k: I now realise that this requirement didn't make it into the final patch, but the extension is still used to identify the provider to upload to, so I see the same issue there.

README.md
ondrejbudai, line 121:
```suggestion
all cloud parameters are provided, e.g.
```

README.md
```diff
@@ -114,6 +114,22 @@ It is possible to generate spdx based SBOM (software bill of materials)
...
+all cloud paramters are provided, e.g.
+```
+$ imge-builder build ami --distro centos-9 \
```
ondrejbudai, line 123:
```suggestion
$ image-builder build ami --distro centos-9 \
```

README.md
```diff
@@ -114,6 +114,22 @@ It is possible to generate spdx based SBOM (software bill of materials)
...
+    --aws-region=us-east-1 \
```
ondrejbudai, line 124: I wonder why you chose `--key=value` notation here when the `--distro` flag one line above uses `--key value`. I think we should be consistent.

README.md
```diff
@@ -114,6 +114,22 @@ It is possible to generate spdx based SBOM (software bill of materials)
...
+	--aws-bucket=example-bucket \
+	--aws-ami-name=my-image-1
+```
+Images can also be uploaded with the `image-builder upload` command
+after they are build.
```
ondrejbudai, line 129:
```suggestion
after they are built.
```

cmd/image-builder/main.go
mvo5: Thank you, that makes sense, I improved the detection based on the actual content, I hope this is what you had in mind. We could also make this explicit and ask the user to provide a type when uploading but that would add another parameter. I can also squash the last commit into the previous one(s) if you want.

README.md
lzap, line 129: This really calls for an explicit `--upload` option for the `ami` argument, just providing AWS credentialy should not automatically upload I think.

cmd/image-builder/upload.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/cheggaaa/pb/v3"
+	"github.com/spf13/cobra"
+
+	"github.com/osbuild/bootc-image-builder/bib/pkg/progress"
+	"github.com/osbuild/images/pkg/cloud"
+	"github.com/osbuild/images/pkg/cloud/awscloud"
+)
+
+var ErrNoUploadConfig = fmt.Errorf("missing or incomplete upload configuration")
```
lzap, line 18:
```suggestion
var ErrNoUploadConfig = errors.New("missing or incomplete upload configuration")
```

ondrejbudai (commented):
Awesome! ❤️ A few typos inline.

I'm also wondering about "upload parameters". Currently, there are two values "hardcoded".

Firstly, the code assumes that the file is always in the "raw" format. However, [AWS supports multiple formats (vmdk, vhd)](https://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-import-snapshot.html#import-snapshot-prerequisites), and I actually think we should investigate using them because they would give us sparse uploads.

Another thing that the code assumes is that the images are hybrid-bootable (with uefi-preference). However, we will soon have support for UKI images that are UEFI-only.

For `image-builder build`, this is not an issue because we know which format we build, and what is its bootmode.

However, for `image-builder upload` I'm afraid we will have to add arguments for setting these two parameters (we can probably guess the format, but the boot mode is just hard to guess). However, this adds a possibility for a PEBKAC.

For a long time, I had this idea of creating a "buildinfo" artifact alongside the built artifact. It would contain machine-readable "high-level" properties of the artifact. Things like boot-mode, and the artifact format can definitely be described in it. Actually, we now have `image-builder describe` command which does a very similar thing.

So I wonder: should we always generate an image description during build, and then be able to consume it in `upload` so we can set all the upload properties instead of requiring a user input?

cmd/image-builder/upload.go
ondrejbudai: I would actually suggest re-thinking this a bit. This error message is not very helpful when a user misses one of the required upload args.

README.md
mvo5: Thanks, would love to hear what @achilleas-k and @ondrejbudai think here, I mostly followed what bib is doing but if this is actually making your life harder and/or something that (most?) users would not expect we should add an explicit upload option. Looking forward for comments :)

README.md
mvo5: Thanks, this is fixed now

cmd/image-builder/main.go
mvo5: For upload this is now an explicit user choice what cloud to use, this should also address the concerns from ondrej that people may upload non-raw images.

mvo5:
[..]
> Firstly, the code assumes that the file is always in the "raw" format. However, [AWS supports multiple formats (vmdk, vhd)](https://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-import-snapshot.html#import-snapshot-prerequisites), and I actually think we should investigate using them because they would give us sparse uploads.
> 
> Another thing that the code assumes is that the images are hybrid-bootable (with uefi-preference). However, we will soon have support for UKI images that are UEFI-only.
> 
> For `image-builder build`, this is not an issue because we know which format we build, and what is its bootmode.
> 
> However, for `image-builder upload` I'm afraid we will have to add arguments for setting these two parameters (we can probably guess the format, but the boot mode is just hard to guess). However, this adds a possibility for a PEBKAC.

Thanks for these thoughts - I think we have some options here
a) just ignore "image-builder upload" for now and allow it only as part of the build 
b) add more options to "image-builder upload" as we need them (e.g. --aws-boot-mode)
c) Add the buildinfo that you describe below and use that (we could still have the flags in b) to override or if buildinfo is missing)

Fwiw, while I like the idea of "image-builder upload" it is not fully clear to me if our users are not already familiar with the "aws" cmdline enough that they really don't need this and just do their usual "aws s3 cp, aws import-image".

> For a long time, I had this idea of creating a "buildinfo" artifact alongside the built artifact. It would contain machine-readable "high-level" properties of the artifact. Things like boot-mode, and the artifact format can definitely be described in it. Actually, we now have `image-builder describe` command which does a very similar thing.
> 
> So I wonder: should we always generate an image description during build, and then be able to consume it in `upload` so we can set all the upload properties instead of requiring a user input?

I like the idea to describe the image more, I would love to talk about the details. I think its very related to https://github.com/osbuild/images/pull/1104

cmd/image-builder/upload.go
mvo5: I did this now. It will give nicer errors - but the downside is that we get the nice errors only in the "image-builder upload" case, when "build" the upload is implicit so any missing parameter will just silently not upload. This is okay (if unexpected) as long as we have "image-builder upload" to fix it after-the-fact.

README.md
mvo5: @ondrejbudai one thing to consider is what I wrote in https://github.com/osbuild/image-builder-cli/pull/119#discussion_r1942524877 - i.e. when we do this implicitly when we have all parameters for an upload we cannot show a nice error if a parameter is missing. That would be a reason for the explicit flag because then we could error if anything is missing.

lzap: I do not have really any logical argument for liking explicit other than it feels more natural to me when there is a specific command to do the same thing:

```
git rebase
git pull --rebase
```

Maybe one argument, it is very common for AWS client tools to accept these values as environment variables, in fact, `bootc-image-builder` does this too. How one can upload an image then all options are specified this way? In the AWS case, there is an option that is required and must be passed regularly, but there might be other clouds that do not have such option. It could be as easy as:

```
export ORACLE_CLOUD_CREDENTIALS=creds.json
cli build --and-upload-to oracle
```

lzap:
I really appreciate the separate upload, because last time I did an AWS bootc build, it failed with "AMI with this name already exists" and I had to do full rebuild.

README.md
```diff
@@ -114,6 +114,22 @@ It is possible to generate spdx based SBOM (software bill of materials)
 documents as part of the build. Just pass `--with-sbom` and
 it will put them into the output directory.
 
+### Cloud integration
+
+When building an image type that can be uploaded to the cloud
+(e.g. an "ami") image-builder will automatically upload if
+all cloud parameters are provided, e.g.
+```
+$ imgae-builder build ami --distro centos-9 \
```
schuellerf, line 123:
```suggestion
$ image-builder build ami --distro centos-9 \
```

cmd/image-builder/main.go
thozza: > Especially since we have plans to let people set the name for the artifact, and I've seen cases where people will choose `.img` or similar.

I'm not sure if it is a good idea to let people specify the extension. My expectation would be that customizing the filename does not affect the extension :innocent:

I tend to agree that checking the extension does not seem to add much if one needs to explicitly specify the cloud to upload to.

thozza (dismissed):
Nice PR.

I've added a comment related about the uploader interface needing to take cloud-specific extra options that are image-type specific.

Moreover, I think that the `--to=` option would make sense also for the `build` command. It could be implicitly set to the default cloud for a given image type, but it would be great to be able to override this. Long time ago, we played with the idea to decouple upload targets from image types, especially for some targets. This makes sense for object stores, such as S3, where you could upload any image type.

cmd/image-builder/upload.go
```diff
@@ -0,0 +1,126 @@
...
+	switch typeOrCloud {
+	case "ami", "aws":
+		return uploaderForCmdAWS(cmd)
+	default:
+		return nil, &UploadTypeUnsupportedError{typeOrCloud}
+	}
+
+}
+
+func uploaderForCmdAWS(cmd *cobra.Command) (cloud.Uploader, error) {
```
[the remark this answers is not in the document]
thozza: I had to have another look at the uploader interface. It is flexible enough, the boot-mode just needs to be pulled out to the uploader options passed when creating one.

thozza (approved):
LGTM

achilleas-k (approved):

cmd/image-builder/upload.go
```diff
@@ -0,0 +1,132 @@
...
+	return fmt.Sprintf("missing upload configuration: %q", e.missing)
+}
+
+type UploadTypeUnsupportedError struct {
+	typ string
+}
+
+func (e *UploadTypeUnsupportedError) Error() string {
+	return fmt.Sprintf("unsupported upload type %q", e.typ)
+}
```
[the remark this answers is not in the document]
mvo5: Nice! This is a great suggestion, I will simplify things accordingly. (We will need ErrPartialUploadConfig and ErrMissingUploadConfig but that should be fine)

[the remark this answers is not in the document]
mvo5: Given that this has 31 comments already and is green I will do it in a followup, hope that is okay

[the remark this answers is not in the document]
lzap: I will do a followup.

PR #133, opened 2025-02-12
main: simplify upload error handling

lzap:
I do not understand what was a point of the error types in the first place when the handling code would just print them. Since strings works great as error values most of the time in Go, it is such easier and cleaner to simply format them correctly when they occur, having them wrapped and let them to bubble up the stack to print them.

This patch fixes that and simplifies the handling code - just return (thus print) any error that occurs, there is no need to do special handling as long as they are correctly formatted and Go error wrapping encourages exactly that.

Additionally, in another commit I am adding `bin/` to the `.gitignore`.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -232,14 +232,13 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	var uploadUnsupported *UploadTypeUnsupportedError
-	var missingUploadConfig *MissingUploadConfigError
 	uploader, err := uploaderFor(cmd, res.ImgType.Name())
-	if err != nil && !errors.As(err, &missingUploadConfig) && !errors.As(err, &uploadUnsupported) {
-		return err
+	if errors.Is(err, ErrUploadTypeUnsupported) || errors.Is(err, ErrUploadConfigNotProvided) {
+		err = nil
 	}
-	if missingUploadConfig != nil && !missingUploadConfig.allMissing {
-		return fmt.Errorf("partial upload config provided: %w", err)
+
+	if err != nil {
+		return err
 	}
 
 	if uploader != nil {
--- a/cmd/image-builder/upload.go
+++ b/cmd/image-builder/upload.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -14,22 +15,14 @@ import (
 	"github.com/osbuild/images/pkg/cloud/awscloud"
 )
 
-type MissingUploadConfigError struct {
-	missing    []string
-	allMissing bool
-}
+// ErrMissingUploadConfig is returned when the upload configuration is missing
+var ErrMissingUploadConfig = errors.New("missing upload configuration")
 
-func (e *MissingUploadConfigError) Error() string {
-	return fmt.Sprintf("missing upload configuration: %q", e.missing)
-}
+// ErrUploadConfigNotProvided is returned when all the upload configuration is missing
+var ErrUploadConfigNotProvided = errors.New("missing all upload configuration")
 
-type UploadTypeUnsupportedError struct {
-	typ string
-}
-
-func (e *UploadTypeUnsupportedError) Error() string {
-	return fmt.Sprintf("unsupported upload type %q", e.typ)
-}
+// ErrUploadTypeUnsupported is returned when the upload type is not supported
+var ErrUploadTypeUnsupported = errors.New("unsupported type")
 
 var awscloudNewUploader = awscloud.NewUploader
 
@@ -73,7 +66,7 @@ func uploaderFor(cmd *cobra.Command, typeOrCloud string) (cloud.Uploader, error)
 	case "ami", "aws":
 		return uploaderForCmdAWS(cmd)
 	default:
-		return nil, &UploadTypeUnsupportedError{typeOrCloud}
+		return nil, fmt.Errorf("%w: %q", ErrUploadTypeUnsupported, typeOrCloud)
 	}
 
 }
@@ -104,10 +97,11 @@ func uploaderForCmdAWS(cmd *cobra.Command) (cloud.Uploader, error) {
 		}
 	}
 	if len(missing) > 0 {
-		return nil, &MissingUploadConfigError{
-			missing:    missing,
-			allMissing: len(missing) == len(requiredArgs),
+		if len(missing) == len(requiredArgs) {
+			return nil, fmt.Errorf("%w: %q", ErrUploadConfigNotProvided, missing)
 		}
+
+		return nil, fmt.Errorf("%w: %q", ErrMissingUploadConfig, missing)
 	}
 
 	return awscloudNewUploader(region, bucketName, amiName, nil)
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ release_artifacts
 *.swp
 dictionary.dic
 __pycache__
+/bin
--- a/cmd/image-builder/main_test.go
+++ b/cmd/image-builder/main_test.go
@@ -343,7 +343,7 @@ func TestBuildIntegrationHappy(t *testing.T) {
 	assert.NoError(t, err)
 
 	// ensure osbuild was run exactly one
-	assert.Equal(t, 1, len(fakeOsbuildCmd.Calls()))
+	require.Equal(t, 1, len(fakeOsbuildCmd.Calls()))
 	osbuildCall := fakeOsbuildCmd.Calls()[0]
 	// --cache is passed correctly to osbuild
 	storePos := slices.Index(osbuildCall, "--store")
--- a/cmd/image-builder/upload_test.go
+++ b/cmd/image-builder/upload_test.go
@@ -98,7 +98,7 @@ func TestUploadCmdlineErrors(t *testing.T) {
 			`missing --to parameter, try --to=aws`,
 		}, {
 			[]string{"--to=aws"},
-			`missing upload configuration: ["--aws-ami-name" "--aws-bucket" "--aws-region"]`,
+			`missing all upload configuration: ["--aws-ami-name" "--aws-bucket" "--aws-region"]`,
 		},
 		{
 			[]string{"--to=aws", "--aws-ami-name=1"},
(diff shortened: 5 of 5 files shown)


cmd/image-builder/main.go
```diff
@@ -213,15 +212,10 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	var uploadUnsupported *UploadTypeUnsupportedError
-	var missingUploadConfig *MissingUploadConfigError
 	uploader, err := uploaderFor(cmd, res.ImgType.Name())
-	if err != nil && !errors.As(err, &missingUploadConfig) && !errors.As(err, &uploadUnsupported) {
+	if err != nil {
```
mvo5, line 216: The reason we have the error types was to distinguish between the following cases:
- image type cannot be uploaded, ignore errors from creating an uploader in this case
- no upload config provided: the user has no intend to upload, ignore the error in this case
- incomplete upload config: the user seems to want to upload, so we need to error to avoid surprises

This still needs to be handled here.

mvo5 (changes requested):
Thanks for the simplification. Unfortunately it went a bit too far, we still need to distinguish the errors (see inline comment). 

I really like the idea of getting rid of the error type and just use annotated ErrFoo vars! Thanks for this!

lzap:
Ok, how about this.

cmd/image-builder/main.go
```diff
@@ -213,13 +213,11 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
 
-	var uploadUnsupported *UploadTypeUnsupportedError
-	var missingUploadConfig *MissingUploadConfigError
 	uploader, err := uploaderFor(cmd, res.ImgType.Name())
-	if err != nil && !errors.As(err, &missingUploadConfig) && !errors.As(err, &uploadUnsupported) {
+	if err != nil && !errors.Is(err, ErrUpload) {
 		return err
 	}
-	if missingUploadConfig != nil && !missingUploadConfig.allMissing {
+	if err != nil && !errors.Is(err, ErrMissingUploadConfig) {
```
mvo5, line 220: This change makes a bunch of the tests fail and it seems its also risky as if we add a new `ErrUpload` there is a risk that this error will get ignored in build. It also makes the test fail because it does not handle the "ErrUploadTypeUnsupported". Something like:
```go
index 69c5d96..ccdfe86 100644
@@ -214,12 +214,9 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
        }
 
        uploader, err := uploaderFor(cmd, res.ImgType.Name())
-       if err != nil && !errors.Is(err, ErrUpload) {
+       if err != nil && !errors.Is(err, ErrUploadTypeUnsupported) && !errors.Is(err, ErrMissingUploadConfigAll) {
                return err
        }
-       if err != nil && !errors.Is(err, ErrMissingUploadConfig) {
-               return fmt.Errorf("partial upload config provided: %w", err)
-       }
 
        if uploader != nil {
                pbar.SetPulseMsgf("Checking cloud access")
```
should make this part work.

cmd/image-builder/upload.go
```diff
@@ -104,10 +100,11 @@ func uploaderForCmdAWS(cmd *cobra.Command) (cloud.Uploader, error) {
 		}
 	}
 	if len(missing) > 0 {
-		return nil, &MissingUploadConfigError{
-			missing:    missing,
-			allMissing: len(missing) == len(requiredArgs),
+		if len(missing) == len(requiredArgs) {
+			return nil, ErrMissingUploadConfigAll
```
mvo5, line 104: This change makes the tests fail and is a UX regression, the tests check that the user gets a message what is missing so that they have the information right away. By just returning a "naked" ErrMissingUploadConfigAll that information is no longer transmitted to the user.

cmd/image-builder/upload.go
```diff
@@ -14,22 +15,17 @@ import (
 	"github.com/osbuild/images/pkg/cloud/awscloud"
 )
 
-type MissingUploadConfigError struct {
-	missing    []string
```
mvo5, line 18: I like the simplicity of the change (a lot!) and I think we should take it. 

However the downside is that a consumer of the error who wants to render the missing arguments in a different way (e.g. by providing a translated error message using gettext or similar) is no longer possible. With the custom type we would either add a Getter or export Missing. With `ErrFoo` and presenting what is missing in a nice and translated way becomes (much) harder.

But this is all theoretical as we do not do i18n in the ibcli yet and we can re-add a type (or use a different approach) if we go down this path.

cmd/image-builder/upload.go
```diff
@@ -14,22 +15,17 @@ import (
...
-	allMissing bool
-}
+// ErrUpload is an error wrapped into all upload errors
+var ErrUpload = errors.New("upload error")
 
-func (e *MissingUploadConfigError) Error() string {
-	return fmt.Sprintf("missing upload configuration: %q", e.missing)
```
mvo5, line 23: This string change will make a bunch of the tests fail because they check for the exact error. We can tweak the errors of course but lets make sure they are still nice and user readable and not too repetitive. I really like go tests that explicitly check the error strings because they tend to get out-of-whack without, e.g.:
```
error depsolving: running osbuild-depsolve-dnf failed:
DNF error occurred: RepoError: There was a problem reading a repository: Failed to download metadata for repo '9828718901ab404ac1b600157aec1a8b19f4b2139e7756f347fb0ecc06c92929' [forced repo#0 xxx.abcdefgh-no-such-host.com/repo: http://xxx.abcdefgh-no-such-host.com/repo]: Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried 
```
is a recent one with "stuttering" (i.e. duplicated strings).

Sadly Go makes this really hard as it is hard to know what a previous error already include, but IME tests help here.

mvo5 (commented):
Thank you! A few comments as the tests are failing (correctly) and the issues are real.

cmd/image-builder/upload.go
lzap: I do not understand how i18n is not possible with using error wrapping. There are several options, replace errors before a new `stringError` (type created by `errors.New`) is created:

```
var ErrMissingUploadConfig = errors.New(_("missing configuration"))
```

Since all strings must be marked for translation anyway, this is the standard way of doing it (underscore is not a valid function in Go I think this is just a pseudo-code). Alternatively, creating a generic translated error type and use that for most errors instead of creating custome errors. This will be pretty easy search and replace operation on `errors.New`. Here is how that could be used (leaving the implementation of `terrors` package on your imagination but can be easily done):

```
var ErrMissingUploadConfig = terrors.New("missing configuration")

// get translated string
ErrMissingUploadConfig.String()

// get original (C) string
ErrMissingUploadConfig.CString()

// wrapping with translating/extracting strings
wrapped := terrors.Errorf("big error: %w", ErrMissingUploadConfig)
```

This would require writing a custom extractor or extending existing one. There is a [drop-in replacement package](https://pkg.go.dev/golang.org/x/text/message) in experimental standard library for string formatting. I like this the most as this could possibly become part of the standard library and it has pretty efficient catalog that is directly linked to the binary, tho, to my test it is pretty verbose and it does not offer extraction function.

For this reason, I like [pure Go gettext implementation](https://github.com/leonelquinteros/gotext), it is straightforward and PO/MO files are universal. The code would look like this:

```
func init() {
 // i18n must be initialized pretty early - preferably with MO files embedded
 // using fs.Embed Go feature - locale change requires app restart
 locale = gotext.NewLocale("/path/to/locales/root/dir", "es_UY")
}

// global function available in the whole app
func L() gotext.Locale {
 return locale
}

var ErrMissingUploadConfig = errors.New(L.Get("missing configuration"))

// etc etc the library provides plurals, fmt-like formatting and variable substitutions
```

cmd/image-builder/upload.go
lzap: Go does not make this hard, if I understand your comment correctly, there is one strict rule (can be enabled via linter even) and if followed you can always unwrap errors via either `Unwrap` or `errors.Is`. The rule is:

```
// never ever return dynamic errors
return errors.New("some random text")
return fmt.Errorf("some random text: random")

// always create an error value
var ErrValue = errors.New("something")

// then either return it or wrap it
return ErrValue
return errors.Join(ErrValue, "my other error")
return fmt.Errorf("something happened: %w", ErrValue)
```

Sadly, many libraries do not follow this rule, specifically AWS SDK (even v2) returns plenty of "dynamic" errors and one need to parse error strings in order to get more details. Yeah.

lzap:
Let me know how you would like me to continue. My opinion is simple - I like simplicity and error values are always cleaner than error types. Future i18n can be done without introducing types per error state, that is way to verbose and way to Java-like.

cmd/image-builder/upload.go
mvo5: Thanks for your interesting idea!

We chatted about this a bit offline - my concerns (and sorry that I did not make this explciit enough) are that "var ErrFoo = " is initialized very early (maybe before gettext is inialized) and only once (so no switch languages later or serving multiple langauges (e.g. in a daemon)) also string concact for i18n is generally tricky.

Having said all that its a big distraction as we do not support i18n right now and when we do we can look into this again :)

cmd/image-builder/upload.go
mvo5: Sorry, maybe I did not express myself clear enough here.

What I mean is that it is hard know in go what context in the error is already provided or not. If I care about nice looking errors that are helpful to the users I might be tempted to write:
```go
func run() error {
	path := "/does/not/exist"
	if _, err := os.Open(path); err != nil {
		return fmt.Errorf("error opening %q: %w", path, err)
	}
	return nil
}

func main() {
	if err := run(); err != nil {
		fmt.Fprintf(os.Stderr, "error: %v\n", err)
		os.Exit(1)
	}
}
```
i.e. provide context what in the open went wrong. But the open error already has this context. So I create "stuttering" which is IME a common problem with errors in go. Without looking at the parent source code (the os package in this cae) I find it hard to know in advance what needs to be added so (again IME) the most reliable way to ensure error message in go are not duplicating or omiting information is to write explicit string compare tests for them.

mvo5: Heh, I think we are talking past each other - sorry that I was not clear enough here. I just agree for unit tests that check error strings. I don't mind which way they are done and I like avoiding the extra complexity of the types.

PR #134, opened 2025-02-12
main: add add `--force-repo` flag

mvo5:
This commit adds an `--force-repo` flag that can be used
to replace all the base repositories with a base url to
a repository. This is useful for testing but also dangerous
as it will not do any checks and happily use a fedora-42 repository
for centos-8 depsolving.

This will make the use-case of the koji builder easier and is
also something that the `build` tool in `images` supports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,55 @@ $ image-builder list-images --output=json
 ]
 ```
 
+## Modifying the set of used repositories
+
+There are various ways to add extra repositories or override the default
+base repositories. Most users will want to use the
+[blueprint systems](https://osbuild.org/docs/user-guide/blueprint-reference/#repositories)
+for this. Repositories that are part of the blueprint will get added
+to the installed image but are not used at build time to install third-party
+packages.
+
+To change repositories during image build time the command line options
+`--datadir`, `--extra-repo` and `--force-repo` can be used. The repositories
+there will only added during build time and will not be available in the
+installed system (use the above blueprint options if that the goal).
+
+Note that both options are targeting advanced users/use-cases and when
+used wrongly can result in failing image builds or non-booting
+systems.
+
+## Using the datadir switch
+
+When using the `--datadir` flag `image-builder` will look into
+the <datadir>/repositories directory for a file called <distro>.json
+that contains the repositories for the <distro>.
+
+This <distro>.json file is a simple architecture-> repositories mapping
+that looks like [this example](https://github.com/osbuild/images/blob/main/data/repositories/centos-10.json).
+
+### Adding extra repositories during the build
+
+To add one or more extra repositories during the build use:
+`--extra-repo <baseurl>`, e.g. `--extra-repo file:///path/to/repo`.
+This will make the content of the repository available during image
+building and the dependency solver will pick packages from there as
+appropriate (e.g. if that repository contains a libc or kernel with a
+higher version number it will be picked over the default
+repositories).
+
+### Overriding the default base repositories during build
+
+To completely replace the default base repositories during a build the
+option `--force-repo=file:///path/to/repos` can be used.
+
+Note that the repositories defined there will be used for all
+dependency solving and there is no safeguards, i.e. one can point to
+a fedora-42 repository url and try to build a centos-9 image type and
+the system will happily try its best (and most likely fail). Use with
+caution.
+
+
 ## FAQ
 
 Q: Does this require a backend.
--- a/cmd/image-builder/repos.go
+++ b/cmd/image-builder/repos.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 
 	"github.com/osbuild/images/data/repositories"
+	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/reporegistry"
 	"github.com/osbuild/images/pkg/rpmmd"
 )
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -189,6 +189,48 @@ $ image-builder list-images --output=json
...
+
+There are various ways to add extra repositories or override the default
+base repositories. Most users will want to use the
+[blueprint systems](https://osbuild.org/docs/user-guide/blueprint-reference/#repositories)
+for this. Repositories that are part of the blueprint will get added
+to the installed image but are not used at build time to install third-party
+packages.
+
+To change repositories during image build time the commandline options
+`--add-repo` and `--override-base-repos-with` can be used. The repositories
```
mvo5, line 202: Happy to discuss the name here: `--override-base-repos-with` -  this is just a strawman :)

thozza: `--force-repo`

supakeen: `--force-repo` is also fine, yes!

mvo5: Thanks, indeed, I misheard during our sync on this I think. Making it symmetric to --extra-repo makes a lot of sense. I updated the PR accordingly. And I like `force-repo`.

README.md
```diff
@@ -189,6 +189,55 @@ $ image-builder list-images --output=json
...
+there will only added during build time and will not be available in the
+installed system (use the above blueprint options if that the goal).
+
+Note that both options are targeting advanced users/use-cases and when
+used wrongly can result in failing image builds or non-booting
+systems.
+
+## Using the datadir switch
+
+When using the `--datadir` flag `image-builder-cli` will look into
```
thozza, line 212: I must admit that I didn't thoroughly check the rest of the docs, but shouldn't we refer to the _binary name_? So `image-builder` vs. `image-builder-cli`.

README.md
```diff
@@ -189,6 +189,55 @@ $ image-builder list-images --output=json
...
+the <datadir>/repositoreis directory for a file called <distro>.json
```
thozza, line 213: typo -> `repositoreis` vs `repositories`

README.md
```diff
@@ -189,6 +189,55 @@ $ image-builder list-images --output=json
...
+that contains the repositories for the <distro>.
+
+This <distro>.json file is a simple architecture-> repositories mapping
+that looks like [this example](https://github.com/osbuild/images/blob/main/data/repositories/centos-10.json).
+
+### Adding extra repositories during the build
+
+To add one or more extra repositories during the build use:
+`--extra-repo <baseurl>`, e.g. `--force-repo file://path/to/repo`.
```
thozza, line 222: The example should use `--extra-repo`, not `--force-repo`.

thozza: Also, the URL is missing one `/` -> `file://path` vs. `file:///path`

README.md
```diff
@@ -189,6 +189,55 @@ $ image-builder list-images --output=json
...
+This will make the content of the repository available during image
+building and the dependency solver will pick packags from there as
+appropriate (e.g. if that repository contains a libc or kernel with a
+higher version number it will be picked over the default
+repositories).
+
+### Overriding the default base repositories during build
+
+To complete replace the default base repositories during a build the
```
thozza, line 231: I think `completely` should be used here instead of `complete`.

thozza (commented):
Thanks, this is nice.

I've added a few comments, but all of them are minor issues.

README.md
```diff
@@ -189,6 +189,55 @@ $ image-builder list-images --output=json
...
+
+There are various ways to add extra repositories or override the default
+base repositories. Most users will want to use the
+[blueprint systems](https://osbuild.org/docs/user-guide/blueprint-reference/#repositories)
+for this. Repositories that are part of the blueprint will get added
+to the installed image but are not used at build time to install third-party
+packages.
+
+To change repositories during image build time the commandline options
+`--datadir`, `--add-repo` and `--force-repo` can be used. The repositories
```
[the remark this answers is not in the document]
mvo5: Silly me, thank you!

README.md
mvo5: Thanks, nice catch! I think just here, fixing

cmd/image-builder/repos.go
```diff
@@ -68,20 +72,15 @@ func newRepoRegistryImpl(dataDir string, extraRepos []string) (*reporegistry.Rep
...
-			}
+	repoConf, err := parseRepoURLs(extraRepos, "extra")
+	if err != nil {
+		return nil, err
+	}
+	for _, repoArchConfigs := range conf {
+		for arch := range repoArchConfigs {
+			archCfg := repoArchConfigs[arch]
+			archCfg = append(archCfg, repoConf...)
+			repoArchConfigs[arch] = archCfg
```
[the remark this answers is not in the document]
thozza: No, you are right. The technically correct thing to do here, without considering cross-arch builds, is to add the repo to the arch that the command targets (build, manifest generation, etc.). For now, this would be the host arch.

I was thinking too ahead with the room for failure because the case that can happen now is no different from the user using repos for the wrong arch.

The code triggered my sense of bugs because while the approach is simple, and I agree that it is harmless within the current implementation constraints, adding extra repos to all architectures is incorrect.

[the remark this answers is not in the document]
mvo5: Sure thing, thank you! I updated the code with a comment and only add to the existing arch.

thozza (approved):
Thank you. Everything looks great :+1:

cmd/image-builder/repos.go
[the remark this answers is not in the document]
thozza: Thank you.

supakeen (approved):
This looks amazing, thank you very much (and @thozza for the thoroug review, nothing left now that I look!).

PR #136, opened 2025-02-18
cmd: run simple riscv64 cross build as part of the integration tests

mvo5:
This branch shows how to cross arch build traditional images. It also allows to skip
the cross-arch build restriction when `IMAGE_BUILDER_EXPERIMENTAL=bootstrap=...` is set.

The not-so-nice part of this work is that we need to pick the the right container via `IMAGE_BUILDER_EXPERIMENTAL=bootstrap=...`  currently. Fortunately almost any upstream container will do as long as it has rpm and setfiles (policycoreutils).

We really should teach images to map containers to buildroots so that we don't need to do this.

Ideally this would be in "images" itself, I guess we will merge ibcli and images eventually anyway.

### DIFF
--- a/test/test_container.py
+++ b/test/test_container.py
@@ -94,3 +94,31 @@ def test_container_with_progress(tmp_path, build_fake_container, progress, needl
     ], text=True)
     assert needle in output
     assert forbidden not in output
+
+
+def test_container_cross_build_riscv64(tmp_path, build_container):
+    # XXX: we should encode this in images proper
+    # note that we need the "fedora-toolbox" container as the regular one
+    # does not contain python3
+    images_experimental_env = "IMAGE_BUILDER_EXPERIMENTAL=bootstrap=ghcr.io/mvo5/fedora-buildroot:41"
+
+    # XXX: only here speed up builds by sharing downloaded stuff
+    os.makedirs("/var/cache/image-builder/store", exist_ok=True)
+    output_dir = tmp_path / "output"
+    output_dir.mkdir()
+    subprocess.check_call([
+        "podman", "run",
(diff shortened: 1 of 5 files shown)


test/test_container.py
```diff
@@ -94,3 +94,32 @@ def test_container_with_progress(tmp_path, build_fake_container, progress, needl
     ], text=True)
     assert needle in output
     assert forbidden not in output
+
+
+def test_container_cross_build_riscv64(tmp_path, build_container):
+    # XXX: we should encode this in images proper
+    # note that we need the "fedora-toolbox" container as the regular one
+    # does not contain python3
+    images_experimental_env = "IMAGES_EXPERIMENTAL=buildroot=fedora.riscv.rocks:3000/davidlt/fedora-toolbox:41"
```
mvo5, line 103: This will need to use ghcr.io/mvo5/fedora-buildroot:41

which is just:
```Containerfile
FROM fedora.riscv.rocks:3000/davidlt/fedora-toolbox:41
RUN dnf install -y selinux-policy-targeted policycoreutils dosfstools
```
#136 opened Feb 18, 2025 by [mvo5](https://github.com/osbuild/image-builder-cli/issues?q=is%3Apr+is%3Aopen+author%3Amvo5)

mvo5: Note that this is an intermediate solution, once the upstream `fedora-toolbox` container installs policycoreutils this container is no longer needed and we can use the upstream one.

mvo5:
Closing for now as https://github.com/osbuild/image-builder-cli/pull/164 also contians a similar (and more comprehensive) test.

PR #146, opened 2025-03-04
chore: bump deps

supakeen:
Bump to latest `images` to make Fedora Minimal work with all the new bits and bobs. Manual PR because the other one was being weird.


bcl (approved):
Looks good to me!

PR #149, opened 2025-03-06
packit: enable epel-9 and epel-10

supakeen:
Let's ship our package in epel-9 and epel-10 as well as all the supported Fedora branches. I've manually created, built, and created updates for these branches already.

I'm not adding automation to the epel-10.0 branch in this PR as we need to probably consider that more carefully. However, I *have* submitted version 13 for epel-10.0 manually.


ondrejbudai (approved):
Ship it!

PR #153, opened 2025-03-10
doc/00-installation: Add link to package details

schuellerf:
There should be a link to the package details, not only the reference to the main fedora page.
Is this link a good choice?


supakeen:
What does linking to `packages.fedoraproject.org` add?

schuellerf:
I am usually somehow lost when it says "our package is in fedora" and then I can't find it -  as it's just the main page.

supakeen (commented):
To me it means that one can install `image-builder` in Fedora.

If it makes it more clear for others what a Fedora package is then linking is fine. (Though that page also doesn't say how to install the package).

bcl (approved):
Looks good to me. It's more useful than just linking to fedora.

PR #157, opened 2025-03-13
tools: provide a `run-in-mock` helper for the koji-plugin

mvo5:
This commit adds a `tools/run-in-mock` script that can be used to wrap running image-builder inside a mock chroot. Without this wrapper bwrap will error because it expects a "/" directory but that is missing in the mock chroots.

Kudos to jlebon and his mock-mount.sh that is available here: https://gist.github.com/jlebon/fb6e7c6dcc3ce17d3e2a86f5938ec033

See also https://github.com/containers/bubblewrap/issues/135

This assume we can just package/run this wrapper from our koji plugin - this is a very simple and straightforward way to solve this, I also did an alternative version that is fully self-contained inside image-builder-cli https://github.com/osbuild/image-builder-cli/compare/main...mvo5:deal-with-mock-self-contained-but-ugly?expand=1 but that contains some tricky (and arguably ugly) re-exec logic that we would only need for this run-in-mock usecase.

### DIFF
--- a/image-builder.spec
+++ b/image-builder.spec
@@ -90,6 +90,8 @@ GOTAGS="exclude_graphdriver_btrfs"
 %install
 install -m 0755 -vd                                 %{buildroot}%{_bindir}
 install -m 0755 -vp %{gobuilddir}/bin/image-builder %{buildroot}%{_bindir}/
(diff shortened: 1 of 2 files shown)


mvo5:
Closing as Simon said he will put this helper directly into the koji plugin (which is fine with me)

PR #158, opened 2025-03-13
ibcli: add new --output-name flag and predictable default names

mvo5:
This commit adds a new `--output-name` flag that will rename the resulting artifact after it was build. All auxillary artifacts like buildlog, sbom etc are also name based on the same basename.

One open question is if this should be "--output-filename" or "--basename".

See also https://github.com/osbuild/images/pull/1039 for how
this could be simpler (especially the fake osbuild).

As a side-effect this will now produce more predictable names by default as well:
```
fedora-42-minimal-raw-zst-x86_64/fedora-42-minimal-raw-zst-x86_64.raw.zst
````

### DIFF
--- a/cmd/image-builder/build.go
+++ b/cmd/image-builder/build.go
@@ -4,31 +4,34 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/osbuild/bootc-image-builder/bib/pkg/progress"
 	"github.com/osbuild/images/pkg/imagefilter"
 )
 
 type buildOptions struct {
-	OutputDir string
-	StoreDir  string
+	OutputDir      string
+	StoreDir       string
+	OutputBasename string
 
 	WriteManifest bool
 	WriteBuildlog bool
 }
 
-func buildImage(pbar progress.ProgressBar, res *imagefilter.Result, osbuildManifest []byte, opts *buildOptions) error {
+func buildImage(pbar progress.ProgressBar, res *imagefilter.Result, osbuildManifest []byte, opts *buildOptions) (string, error) {
 	if opts == nil {
 		opts = &buildOptions{}
 	}
 
+	basename := basenameFor(res, opts.OutputBasename)
 	if opts.WriteManifest {
-		p := filepath.Join(opts.OutputDir, fmt.Sprintf("%s.osbuild-manifest.json", outputNameFor(res)))
+		p := filepath.Join(opts.OutputDir, fmt.Sprintf("%s.osbuild-manifest.json", basename))
 		if err := os.MkdirAll(filepath.Dir(p), 0755); err != nil {
-			return err
+			return "", err
 		}
 		if err := os.WriteFile(p, osbuildManifest, 0644); err != nil {
-			return err
+			return "", err
 		}
 	}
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by supakeen — APPROVED** [review]
I ❤️ this, thank you. I'm perfectly fine with `--output-name` as that name makes sense to me, it only gets murky when *combined* with `--output-dir`, but not much.

---

For other readers, this would hugely simplify renaming and other bits and bobs I currently would need to do for `koji` integration.

File: cmd/image-builder/build.go [hunk 1]
```diff
@@ -4,31 +4,34 @@ import (
...
 	WriteBuildlog bool
 }
 
-func buildImage(pbar progress.ProgressBar, res *imagefilter.Result, osbuildManifest []byte, opts *buildOptions) error {
+func buildImage(pbar progress.ProgressBar, res *imagefilter.Result, osbuildManifest []byte, opts *buildOptions) (string, error) {
 	if opts == nil {
 		opts = &buildOptions{}
 	}
 
+	basename := basenameFor(res, opts.OutputBasename)
```
supakeen, line 27: This seems to slightly mishandle basenames which already contain dots in them, for example:

```
user@muja image-builder-cli € ll output 
total 488396
-rw-r--r--. 1 root root    315185 Mar 14 12:21 Fedora-Minimal-42-20250314.buildroot-build.spdx.json
-rw-r--r--. 1 root root    750529 Mar 14 12:21 Fedora-Minimal-42-20250314.image-os.spdx.json
-rw-r--r--. 1 root root     79595 Mar 14 12:23 Fedora-Minimal-42-20250314.n.1.buildlog
-rw-r--r--. 1 root root    153945 Mar 14 12:21 Fedora-Minimal-42-20250314.n.1.osbuild-manifest.json
-rw-r--r--. 1 root root 498810732 Mar 14 12:23 Fedora-Minimal-42-20250314.n.1.raw.xz
```

supakeen (changes requested):
Found a tiny bug.

cmd/image-builder/manifest.go
```diff
@@ -52,11 +55,12 @@ func generateManifest(dataDir string, extraRepos []string, img *imagefilter.Resu
 		RpmDownloader: opts.RpmDownloader,
 	}
 	if opts.WithSBOM {
-		outputDir := opts.OutputDir
-		if outputDir == "" {
-			outputDir = outputNameFor(img)
-		}
+		outputDir := basenameFor(img, opts.OutputDir)
+		overrideSBOMBase := strings.SplitN(opts.OutputFilename, ".", 2)[0]
```
supakeen, line 59: This seems to slightly mishandle basenames which already contain dots in them, for example:

```
user@muja image-builder-cli € sudo ./image-builder build --output-name Fedora-Minimal-42-20250314.n.1 ... minimal-raw
# ...
user@muja image-builder-cli € ll output 
total 488396
-rw-r--r--. 1 root root    315185 Mar 14 12:21 Fedora-Minimal-42-20250314.buildroot-build.spdx.json
-rw-r--r--. 1 root root    750529 Mar 14 12:21 Fedora-Minimal-42-20250314.image-os.spdx.json
-rw-r--r--. 1 root root     79595 Mar 14 12:23 Fedora-Minimal-42-20250314.n.1.buildlog
-rw-r--r--. 1 root root    153945 Mar 14 12:21 Fedora-Minimal-42-20250314.n.1.osbuild-manifest.json
-rw-r--r--. 1 root root 498810732 Mar 14 12:23 Fedora-Minimal-42-20250314.n.1.raw.xz
```

Note the `buildroot-build.spdx.json` and `image-os.spdx.json` files have dropped the `.n.1` part of the basename.

supakeen (approved):
Fixed a tiny bug. All seems good now (but weird to approve a PR with my own commits).

cmd/image-builder/build.go
```diff
@@ -38,16 +41,32 @@ func buildImage(pbar progress.ProgressBar, res *imagefilter.Result, osbuildManif
...
+	// similar code like this.
+	pipelineDir := filepath.Join(opts.OutputDir, res.ImgType.Exports()[0])
+	srcName := filepath.Join(pipelineDir, res.ImgType.Filename())
+	imgExt := strings.SplitN(res.ImgType.Filename(), ".", 2)[1]
+	dstName := filepath.Join(opts.OutputDir, fmt.Sprintf("%s.%v", basename, imgExt))
+	if err := os.Rename(srcName, dstName); err != nil {
+		return "", fmt.Errorf("cannot rename artifact to final name: %w", err)
+	}
+	// best effort, remove the now empty pipeline export dir from osbuild
+	_ = os.Remove(pipelineDir)
```
achilleas-k, line 69: Wont `os.Rename()` fail if the source and destination are on different filesystems?  Wouldn't it be bad if it fails and also deletes the source?  Could instead let the user know that the move failed but the artifact is available at `pipelineDir`.

achilleas-k: > Wont `os.Rename()` fail if the source and destination are on different filesystems?

To be clear, I bring this up because I think it might be a very common use case.

mvo5: Yes, os.Rename() will fail if src and destination are on different file-systems. This (optimistically) assumes that the "export" directory under outputDir is on the same filesystem and it will result in confusing errors if not. The only error condition I can think of is if someone has symlinks in $outputDir that matches the name of the exported pipeline, e.g. $outputDir/ec2 -> /mnt/other-disk" (which is hopefully rare but https://xkcd.com/1172/)

I can look into fixing this in osbuild to have an option there to not hardcode the output dir to the export filename or alternatively instead of exporting into a subdir prefix the pipeline-name if the concern is clashes.

achilleas-k: ah, I think I misunderstood the meaning of `pipelineDir`.

achilleas-k (approved):

PR #164, opened 2025-03-21
main: automatically cross build when --arch <foreign> is passed

mvo5:
This PR enables automatic cross architecture building when `build --arch <foreign>` is passed. It uses the new generalized bootstrap buildroot feature of images.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/osbuild/bootc-image-builder/bib/pkg/progress"
 	"github.com/osbuild/images/pkg/arch"
-	"github.com/osbuild/images/pkg/experimentalflags"
 	"github.com/osbuild/images/pkg/imagefilter"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/ostree"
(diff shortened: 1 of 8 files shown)


cmd/image-builder/main.go
```diff
@@ -82,7 +81,7 @@ func ostreeImageOptions(cmd *cobra.Command) (*ostree.ImageOptions, error) {
 	}, nil
 }
 
-func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []string, w io.Writer, archChecker func(string) error) (*imagefilter.Result, error) {
+func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []string, w io.Writer, needBootstrap func(string) bool) (*imagefilter.Result, error) {
```
achilleas-k, line 84: Having  a callback for this rather than an argument seems strange to me.  I realise boolean args are sometimes hard to read from the caller side, but is that the only reason?

Playing with this a bit, I suppose having a simple bool would change
```go
	res, err := cmdManifestWrapper(pbar, cmd, args, &mf, func(archStr string) bool {
		return archStr != arch.Current().String()
	})
```
to
```go
    archStr, err := cmd.Flags().GetString("arch")
	if archStr == "" {
		archStr = arch.Current().String()
	}
	res, err := cmdManifestWrapper(pbar, cmd, args, &mf, archStr != arch.Current().String())
```
which seems more verbose and would require pulling `"arch"` from the flags twice.

Alternatively, the option could be made to mean "add bootstrap pipeline if needed", which would change
```go
	if needBootstrap != nil {
		opts.UseBootstrapContainer = needBootstrap(img.Arch.Name())
	}
```
to
```go
   opts.UseBootstrapContainer = useBootstrapIfNeeded && archStr != arch.Current().String()
```

achilleas-k (approved):
LGTM.  One comment in-line but feel free to ignore or leave for another time.

cmd/image-builder/main.go
mvo5: Thanks for your thoughts on this - I agree that it is currently a bit cumbersome with the callback. I will open a followup based on your suggestion(s).

PR #167, opened 2025-03-25
main: tweak how ibcli determines if bootstraping is needed

mvo5:
This simplifies the way that ibcli determines if the manifest needs a bootstrap stage or not. The logic is simple:
- when doing `image-builder manifest` we never include one
- when doing `image-builder build --arch` we add one if the target arch and host arch differ

Instead of passing a callback this is now just a simple bool option. Thanks to Achilleas for suggesting this.


achilleas-k (approved):
That was quick!

PR #168, opened 2025-03-25
README: document that cross building works

mvo5:
Document how cross building works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,20 @@ $ sudo image-builder build qcow2 --blueprint ./config.toml --distro centos-9
 ...
 ```
 
+### Cross architecture building
+
+When `qemu-user-static` is installed images can be build for foreign
+architectures. To do this, pass `--arch`, e.g.:
+
+```console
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,6 +121,20 @@ $ sudo image-builder build qcow2 --blueprint ./config.toml --distro centos-9
 ...
 ```
 
+### Cross architecture building
+
+When `qemu-user-static` is installed images can be build for foreign
+architectures. To do this, just pass `--arch`, e.g.:
```
supakeen, line 127: Let's remove the "just".

README.md
```diff
@@ -121,6 +121,20 @@ $ sudo image-builder build qcow2 --blueprint ./config.toml --distro centos-9
...
+
+```console
+$ sudo image-builder build --arch=riscv64 minimal-raw --distro fedora-41
+```
+building is about 8x-10x slower than native building but still fast
+enough to be usable.
+
+Note that feature is considered experimental currently.
```
supakeen, line 135: "this feature".

Issue #114, opened 2025-01-31
Enable caching for the build pipeline


ondrejbudai:
It seems like `image-builder-cli` doesn't enable any caching, see the output from `ps`:

```
/usr/bin/osbuild --store /var/cache/image-builder/store --output-directory fedora-41-qcow2-x86_64 - --export qcow2
```

Can we at least enable caching for the build pipeline with some reasonable default cache size (dunno, 20 GiB)? We can iterate from there. :)

mvo5:
We will need https://github.com/osbuild/bootc-image-builder/pull/812 as a prepreq I think.

PR #176, opened 2025-03-31
main: allow seed setting

supakeen:
Allow users to define the seed that's used for manifest generation. Regenerating an image with a given seed leads to the same manifest (provided depsolving does the same).

The seed is normally mostly used to generate random filesystem UUIDs.

This will need a bunch of documentation in a follow up since the use cases are meant to be advanced but it can really speed up multi-type and rebuilds of the same image.

---

Note that this is part of an ongoing effort, with this change users can opt for the same manifests being generated (barring depsolve changes). This stops the invalidation of the `os` pipeline, which contains filesystem UUIDs. This can then, in a followup, lead to us check pointing more; allowing us to re-use more.

Related to #114, probably only really useful after https://github.com/osbuild/bootc-image-builder/pull/812

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -151,6 +151,14 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 	if err != nil {
 		return nil, err
 	}
+	var customSeed *int64
+	if cmd.Flags().Changed("seed") {
+		seedFlagVal, err := cmd.Flags().GetInt64("seed")
+		if err != nil {
+			return nil, err
+		}
+		customSeed = &seedFlagVal
+	}
 	// no error check here as this is (deliberately) not defined on
 	// "manifest" (if "images" learn to set the output filename in
 	// manifests we would change this
@@ -189,6 +197,7 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 		Ostree:         ostreeImgOpts,
 		RpmDownloader:  rpmDownloader,
 		WithSBOM:       withSBOM,
+		CustomSeed:     customSeed,
 
 		ForceRepos: forceRepos,
 	}
@@ -398,6 +407,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 		Hidden:       true,
 	}
 	manifestCmd.Flags().String("blueprint", "", `filename of a blueprint to customize an image`)
+	manifestCmd.Flags().Int64("seed", 0, `rng seed, some values are derived randomly, pinning the seed allows more reproducibility if you need it. must be an integer. only used when changed.`)
 	manifestCmd.Flags().String("arch", "", `build manifest for a different architecture`)
 	manifestCmd.Flags().String("distro", "", `build manifest for a different distroname (e.g. centos-9)`)
 	manifestCmd.Flags().String("ostree-ref", "", `OSTREE reference`)
--- a/test/test_container.py
+++ b/test/test_container.py
@@ -122,3 +122,37 @@ def test_container_cross_build(tmp_path, build_container, arch):
         f"--arch={arch}",
     ], text=True)
     assert os.path.exists(output_dir / f"fedora-41-container-{arch}.tar")
+
+
+@pytest.mark.parametrize("use_seed_arg", [False, True])
+@pytest.mark.skipif(os.getuid() != 0, reason="needs root")
+def test_container_manifest_seeded_is_the_same(build_container, use_seed_arg):
+    manifests = set()
+
+    cmd = [
+        "podman", "run",
+        "--privileged",
+        build_container,
+        "manifest",
+        "--distro", "centos-9",
+        "minimal-raw",
+    ]
+
+    if use_seed_arg:
+        cmd.extend(["--seed", "0"])
+
+    for _ in range(3):
+        p = subprocess.run(
+            cmd,
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE)
+
+        manifests.add(p.stdout)
+
+    # verify all calls with the same seed generated the same manifest
+    if use_seed_arg:
+        assert len(manifests) == 1
+    else:
+        print(cmd)
+        assert len(manifests) == 3
--- a/cmd/image-builder/manifest.go
+++ b/cmd/image-builder/manifest.go
@@ -24,6 +24,7 @@ type manifestOptions struct {
 	Ostree         *ostree.ImageOptions
 	RpmDownloader  osbuild.RpmDownloader
 	WithSBOM       bool
+	CustomSeed     *int64
 
 	ForceRepos            []string
 	UseBootstrapContainer bool
@@ -55,6 +56,7 @@ func generateManifest(dataDir string, extraRepos []string, img *imagefilter.Resu
 		Output:                output,
 		RpmDownloader:         opts.RpmDownloader,
 		UseBootstrapContainer: opts.UseBootstrapContainer,
+		CustomSeed:            opts.CustomSeed,
 	}
 	if opts.WithSBOM {
 		outputDir := basenameFor(img, opts.OutputDir)


cmd/image-builder/main.go
```diff
@@ -151,6 +152,10 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 	if err != nil {
 		return nil, err
 	}
+	seedValue, err := cmd.Flags().GetString("seed")
```
mvo5, line 155: Could we just use an integer seed here? I.e. just `GetInt64()`? Having it as a int would have the advantage of:
a) removes some code here as we don't need to hash
b) we could make the seed for manifests available somehoe (print, put into the buildlog etc) even if "--seed" is not provided and users could use that int then as input here

test/test_container.py
```diff
@@ -122,3 +122,25 @@ def test_container_cross_build(tmp_path, build_container, arch):
         f"--arch={arch}",
     ], text=True)
     assert os.path.exists(output_dir / f"fedora-41-container-{arch}.tar")
+
+
+@pytest.mark.skipif(os.getuid() != 0, reason="needs root")
+def test_container_manifest_seeded_is_the_same(build_container):
+    data = [
```
mvo5, line 129: Thanks for this test! Its concise and uses some nice python features. But (at least) personally I fine the reverse style of list comprehensive harder to follow, maybe something more linear like:
```python
def test_container_manifest_seeded_is_the_same(build_container):
    osbuild_manifests = set()
    for _ in range(3):
        p = subprocess.run([
            "podman", "run",
            "--privileged",
            "-v", "/var/cache/image-builder/store:/var/cache/image-builder/store",
            build_container,
            "manifest",
            "--distro", "centos-9",
            "--seed", "hello",
            "minimal-raw",
        ], check=True, capture_output=True)
        osbuild_manifests.add(p.stdout)
    # verify all stdouts for the 3 calls were the same
    assert len(osbuild_manifests) == 1
```

mvo5 (commented):
Thanks! I like this feature (and the direction) - some ideas/suggestions inline for your consideration.

cmd/image-builder/main.go
supakeen: Sure, I can make it a direct integer instead. I picked hashing because I felt it gives a nicer interface to the user.

test/test_container.py
supakeen: Sure, if this is more readable to you then I'll change it to be this.

cmd/image-builder/main.go
```diff
@@ -182,13 +186,20 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 		return nil, fmt.Errorf("image %q has multiple exports: this is current unsupport: please report this as a bug", basenameFor(img, ""))
 	}
 
+	var customSeed *int64
+	if seedValue != 0 {
```
mvo5, line 190: I think we can use `cmd.Flags().Changed("seed")` here and with that `0` is no longer special, something like:
```go
diff --git a/cmd/image-builder/main.go b/cmd/image-builder/main.go
index eccfa60..01921fa 100644
@@ -151,9 +151,13 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 	if err != nil {
 		return nil, err
 	}
-	seedValue, err := cmd.Flags().GetInt("seed")
-	if err != nil {
-		return nil, err
+	var customSeed *int64
+	if cmd.Flags().Changed("seed") {
+		seedFlagVal, err := cmd.Flags().GetInt64("seed")
+		if err != nil {
+			return nil, err
+		}
+		customSeed = &seedFlagVal
 	}
 	// no error check here as this is (deliberately) not defined on
 	// "manifest" (if "images" learn to set the output filename in
@@ -186,12 +190,6 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 		return nil, fmt.Errorf("image %q has multiple exports: this is current unsupport: please report this as a bug", basenameFor(img, ""))
 	}
 
-	var customSeed *int64
-	if seedValue != 0 {
-		seedAs64 := int64(seedValue)
-		customSeed = &seedAs64
-	}
-
 	opts := &manifestOptions{
 		OutputDir:      outputDir,
 		OutputFilename: outputFilename,
@@ -409,7 +407,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 		Hidden:       true,
 	}
 	manifestCmd.Flags().String("blueprint", "", `filename of a blueprint to customize an image`)
-	manifestCmd.Flags().Int("seed", 0, `rng seed, some values are derived randomly, pinning the seed allows more reproducibility if you need it. must be an integer. a value of 0 means a random seed`)
+	manifestCmd.Flags().Int64("seed", 0, `rng seed, some values are derived randomly, pinning the seed allows more reproducibility if you need it. must be an integer. a value of 0 means a random seed`)
 	manifestCmd.Flags().String("arch", "", `build manifest for a different architecture`)
 	manifestCmd.Flags().String("distro", "", `build manifest for a different distroname (e.g. centos-9)`)
 	manifestCmd.Flags().String("ostree-ref", "", `OSTREE reference`)
diff --git a/test/test_container.py b/test/test_container.py
index 5d76d6b..e2c3cb4 100644
--- a/test/test_container.py
+++ b/test/test_container.py
@@ -124,21 +124,24 @@ def test_container_cross_build(tmp_path, build_container, arch):
     assert os.path.exists(output_dir / f"fedora-41-container-{arch}.tar")
 
 
+@pytest.mark.parametrize("use_seed_arg", [False, True])    
-def test_container_manifest_seeded_is_the_same(build_container):
+def test_container_manifest_seeded_is_the_same(build_container, use_seed_arg):
     manifests = set()
 
+    cmd = [
+        "podman", "run",
+        "--privileged",
+        build_container,
+        "manifest",
+        "--distro", "centos-9",
+        "minimal-raw",
+    ]
+    if use_seed_arg:
+        cmd.extend("--seed", "0")
+
     for _ in range(3):
-        p = subprocess.run(
-            [
-                "podman", "run",
-                "--privileged",
-                build_container,
-                "manifest",
-                "--distro", "centos-9",
-                "--seed", "1",
-                "minimal-raw",
-            ],
+        p = subprocess.run(cmd,
             check=True,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE)
@@ -146,4 +149,7 @@ def test_container_manifest_seeded_is_the_same(build_container):
         manifests.add(p.stdout)
 
     # verify all calls with the same seed generated the same manifest
-    assert len(manifests) == 1
+    if use_seed_arg:
+        assert len(manifests) == 1
+    else:
+        assert len(manifests) == 3

```

mvo5 (commented):
Thank, I like it! Another small suggestion inline for the cobra use. The test change is a bit optional, it depends on how long it takes (it seems its fast) but in general its nice to best both the yes and the no case.

cmd/image-builder/main.go
supakeen: Sure.

mvo5 (commented):
Thanks! Looks really nice now, I'm :+1: for merging, let's just do a final squash of the commits into one or two (I think one is fine but have no strong preference, I know some ppl like to have the test in their own commits).

supakeen:
> Thanks! Looks really nice now, I'm 👍 for merging, let's just do a final squash of the commits into one or two (I think one is fine but have no strong preference, I know some ppl like to have the test in their own commits).

Done.

test/test_container.py
```diff
@@ -122,3 +122,37 @@ def test_container_cross_build(tmp_path, build_container, arch):
...
+    ]
+
+    if use_seed_arg:
+        cmd.extend(["--seed", "0"])
+
+    for _ in range(3):
+        p = subprocess.run(
+            cmd,
+            check=True,
+            stdout=subprocess.PIPE,
```
mvo5, line 148: (super nitick) any deeper reason we don't just use `capture_output=True` here`Or just slipped though? I don't really mind fwiw

supakeen: This way `stderr` can be captured by pytest.

achilleas-k (approved):

PR #177, opened 2025-03-31
main: rename `list-images` to `list` (with alias for the old names)

### LINKED ISSUE #78: Rename `list-images` to `list`.
More general naming shenanigans. Since all of our subcommands generally will work on images I see no good reason to have this command contain images in the name. If we want to have non-image commands we have a few options:

We could have:

1. `image-builder build [image-type]`
2. `image-builder describe [image-type]` (see #9).
3. `image-builder list`

Or alternatively we could have `-image` in all of them:

1. `image-builder build-image [image-type]`
2. `image-builder describe-image [image-type]`
3. `image-builder list-image`

Or alternatively they can all exist under an `image` subcommand:

1. `image-builder image build [image-type]`
2. `image-builder image describe [image-type]`
3. `image-builder image list`

Which one do you prefer?

supakeen:
Rename the `list-images` command to `list`. We don't have `-image(s)` in our other subcommands so this is for consistency.

We keep a `list-images` alias behind for compatibility reasons.

---

Closes #78, name can be discussed but the initial issue didn't garner attention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ $ sudo dnf install osbuild osbuild-depsolve-dnf
 
 To see the list of buildable images run:
 ```console
-$ image-builder list-images
+$ image-builder list
 ...
 centos-9 type:qcow2 arch:x86_64
 ...
@@ -161,7 +161,7 @@ after they are built.
 
 When listing images, it is possible to filter:
 ```console
-$ image-builder list-images --filter ami
+$ image-builder list --filter ami
 ...
 centos-9 type:ami arch:x86_64
 ...
@@ -171,7 +171,7 @@ rhel-10.0 type:ami arch:aarch64
 ```
 or be more specific
 ```console
-$ image-builder list-images --filter "arch:x86*" --filter "distro:*centos*"
+$ image-builder list --filter "arch:x86*" --filter "distro:*centos*"
 centos-9 type:ami arch:x86_64
 ...
 centos-9 type:qcow2 arch:x86_64
@@ -188,7 +188,7 @@ The following filters are currently supported, shell-style globbing is supported
 
 The text format can also be switched, supported are "text", "json":
 ```console
-$ image-builder list-images --format=json
+$ image-builder list --format=json
 [
   {
     "distro": {
--- a/cmd/image-builder/filters.go
+++ b/cmd/image-builder/filters.go
@@ -42,7 +42,7 @@ func getOneImage(distroName, imgTypeStr, archStr string, repoOpts *repoOptions) 
 		return nil, err
 	}
 	// strip prefixes to make ib copy/paste friendly when pasting output
-	// from "list-images"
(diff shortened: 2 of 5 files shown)


bcl (dismissed):
Looks good to me

mvo5:
I'm not 100% on-board (but also not 100% opposed) (but maybe just because of habbit by now).

The argument against would be that the (mandatory) argument of build gives the information away `image-builder build <image-type>` but for `list` there is no argument. But if we do it we should also do it for describe-image -> describe (just as outlined in the issue).

supakeen:
> I'm not 100% on-board (but also not 100% opposed) (but maybe just because of habbit by now).
> 
> The argument against would be that the (mandatory) argument of build gives the information away `image-builder build <image-type>` but for `list` there is no argument. But if we do it we should also do it for describe-image -> describe (just as outlined in the issue).

Under that reasoning `describe` follows the same pattern as `build` (it takes an image as its direct argument) so that one would be even more consistent! I've added a commit that also renames `describe` (and leaves `describe-image` as its alias).

supakeen:
FWIW, I'm fine to drop the `list-images` commit here if that's the controversial bit though it would be the "odd one out". My feeling is that *all* these subcommands already implicitly interact on images, mostly because I can't really think of a different meaning for `image-builder list` than "lists images".

mvo5:
> FWIW, I'm fine to drop the `list-images` commit here if that's the controversial bit though it would be the "odd one out". My feeling is that _all_ these subcommands already implicitly interact on images, mostly because I can't really think of a different meaning for `image-builder list` than "lists images".

I mean, we could simply have two PRs, merge the uncontroversial one, then ponder a bit about the other one (and hear more feedback). 

(unrelated - I'm also contemplating making `describe` no longer hidden, with the `partition_table` addition and the (soon hopefully) converage of the output with our image type yaml we hopefully it gets more and more non-specialized I feel)

supakeen:
> > FWIW, I'm fine to drop the `list-images` commit here if that's the controversial bit though it would be the "odd one out". My feeling is that _all_ these subcommands already implicitly interact on images, mostly because I can't really think of a different meaning for `image-builder list` than "lists images".
> 
> I mean, we could simply have two PRs, merge the uncontroversial one, then ponder a bit about the other one (and hear more feedback).
> 
> (unrelated - I'm also contemplating making `describe` no longer hidden, with the `partition_table` addition and the (soon hopefully) converage of the output with our image type yaml we hopefully it gets more and more non-specialized I feel)

Split into #186 and dropped the commit from here.

bcl (approved):
Still ok with me :)

mvo5:
Thanks, I'm fine with the rename then (still not 100% but the old name is also not 100%)  there is probably nothing 100% fitting and both apt and dnf have list to show their packages so this seems analogous (enough).

Issue #103, opened 2025-01-27
Running without `--privileged` returns a mount error


ondrejbudai:
```console
$ sudo podman run --rm ghcr.io/osbuild/image-builder-cli
mount: /run/osbuild: permission denied.
       dmesg(1) may have more information after failed mount system call.
```

I expect the container:

- not to require `--privileged` for showing help
- privileged commands (`build`) to show a helpful error

mvo5:
We probably need to do what we do in bib and move the container setup into the go code itself so that it has context what it needs (c.f. bibs `internal/setup/setup.go` code)

Issue #76, opened 2025-01-19
Handle permissions better


supakeen:
Image Builder requires root permissions for many tasks (but not all). This is due to needing root permissions to do any disk mounting.

However, `image-builder-cli` *always* needs root permissions due to trying to create a directory in `/var/cache`.

If we're going down that route then perhaps we should add a `geteuid()` check and exit earlier. Right now it takes (on my system) 10-15 seconds before we get to the point of error:

```
user@muja ~ € image-builder build iot-commit
No distro name specified, selecting "fedora-41" based on host, use --distro to overrideWARN[0000] Failed to load consumer certs: no consumer key found 
Traceback (most recent call last):
  File "/usr/bin/osbuild", line 33, in <module>
    sys.exit(load_entry_point('osbuild==138', 'console_scripts', 'osbuild')())
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/osbuild/main_cli.py", line 171, in osbuild_cli
    with ObjectStore(args.store) as object_store:
         ~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/osbuild/objectstore.py", line 325, in __init__
    os.makedirs(self.store, exist_ok=True)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen os>", line 217, in makedirs
  File "<frozen os>", line 227, in makedirs
PermissionError: [Errno 13] Permission denied: '/var/cache/image-builder'
error: running osbuild failed: exit status 1
```

Alternatively, we might want to tag the various image types with *if* they require root permissions and handle the cache differently.

PR #178, opened 2025-03-31
import: `setup`, `util`, `podmanutil` from `bootc-image-builder`

supakeen:
To implement #76 and #103 we would like to re-use `bootc-image-builder'`s `setup` module. Let's import it here (and make `bootc-image-builder` import it from here afterwards).

---

Note that I took over all history but there is a single commit on top that puts the files in the correct places since I don't quite understand if I'm able to do that rename during the import.


mvo5 (commented):
This looks fine to me, I think its good, maybe a quick double check from @achilleas-k would be nice but afaict the history import looks fine.

achilleas-k (approved):
LGTM.

Question for the future: Most of these checks are used in bootc-image-builder unconditionally because bib is meant to always run from a container.  Are we going to have variants of these for non-containerised scenarios (checking if we have root for example), or will these only be useful for the containerised ibcli?

supakeen:
> LGTM.
> 
> Question for the future: Most of these checks are used in bootc-image-builder unconditionally because bib is meant to always run from a container. Are we going to have variants of these for non-containerised scenarios (checking if we have root for example), or will these only be useful for the containerised ibcli?

In my opinion we should cover both use cases, as follow-ups.

PR #179, opened 2025-03-31
import: `progress` from `bootc-image-builder`

supakeen:
As preparation for rolling `bootc-image-builder` into `image-builder` let's import its progress implementation.

This PR has two commits on top, one that moves the imported files to the proper path within this repository, and another that changes import paths to refer there.

Will need a follow-up in `bootc-image-builder` to reverse the imports after this merges.


achilleas-k (approved):
Great.  LGTM.

(I didn't read every commit line-by-line since they're imported, but the commit messages look good)

PR #182, opened 2025-04-01
progress: set --cache-max-size in osbuild

supakeen:
This commit allows controlling the `osbuild --cache-max-size` option. By default it will set the cache to 20GiB but allows this to be controlled by the user.

Thanks to Simon for raising this.

---

Commit imported from https://github.com/osbuild/bootc-image-builder/pull/812


mvo5 (approved):
Thank you!

PR #183, opened 2025-04-01
main: add base partition table to `describe-image`

mvo5:
[draft as it needs https://github.com/osbuild/images/pull/1376 first and once that is in we need to drop the "go mod edit -replace"  commit, also we need to think if that should be `partition_table` or `base_partition_table` in the output]

This commit adds the (base) partition table to the `describe-image`
command. It needs https://github.com/osbuild/images/pull/1376

It currently looks like:
```yaml
$ image-builder describe-image qcow2 --distro fedora-41
@WARNING - the output format is not stable yet and may change
distro: fedora-41
...
partition_table:
  uuid: D209C89E-EA5E-4FBD-B161-B461CCE297E0
  type: gpt
  partitions:
    - size: 1048576
      type: 21686148-6449-6E6F-744E-656564454649
      bootable: true
      uuid: FAC7F1FB-3E8D-4137-A512-961DE09A5549
    - size: 209715200
      type: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
      uuid: 68B2905B-DF3E-4FB3-80FA-49D1E773AA33
      payload:
        type: vfat
        uuid: 7B77-95E7
        label: EFI-SYSTEM
        mountpoint: /boot/efi
        fstab_options: defaults,uid=0,gid=0,umask=077,shortname=winnt
        fstab_passno: 2
    - size: 524288000
      type: 0FC63DAF-8483-4772-8E79-3D69D8477DE4
      uuid: CB07C243-BC44-4717-853E-28852021225B
      payload:
        type: ext4
        label: boot
        mountpoint: /boot
        fstab_options: defaults
    - size: 2147483648
      type: 0FC63DAF-8483-4772-8E79-3D69D8477DE4
      uuid: 6264D520-3FB9-423F-8AB8-7A0A8E3D3562
      payload:
        type: ext4
        label: root
        mountpoint: /
        fstab_options: defaults
```

### DIFF
--- a/cmd/image-builder/describeimg.go
+++ b/cmd/image-builder/describeimg.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"slices"
@@ -8,7 +9,9 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/osbuild/images/pkg/disk"
 	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/distro/defs"
 	"github.com/osbuild/images/pkg/imagefilter"
 )
 
@@ -31,6 +34,8 @@ type describeImgYAML struct {
 	BuildPipelines   []string                 `yaml:"build_pipelines"`
 	PayloadPipelines []string                 `yaml:"payload_pipelines"`
 	Packages         map[string]*packagesYAML `yaml:"packages"`
+
+	PartitionTable *disk.PartitionTable `yaml:"partition_table,omitempty"`
 }
 
 type packagesYAML struct {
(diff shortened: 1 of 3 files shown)


cmd/image-builder/describeimg.go
```diff
@@ -31,6 +34,8 @@ type describeImgYAML struct {
 	BuildPipelines   []string                 `yaml:"build_pipelines"`
 	PayloadPipelines []string                 `yaml:"payload_pipelines"`
 	Packages         map[string]*packagesYAML `yaml:"packages"`
+
+	PartitionTable *disk.PartitionTable `yaml:"partition_table,omitempty"`
```
mvo5, line 38: I guess one open question is if this should be `PartitionTable` or `BasePartitionTable`

achilleas-k: I think plain `PartitionTable` is enough.  I think it's clear that what we're displaying is the default configuration for the image in all respects.  Packages, hostname, locales, and all other configurable properties are displayed as their _base_ values, even though they can be changed but we don't specify that it's the "base package set", do we?

This also makes me thing that an interesting idea might be to support `describe-image` with a blueprint that modifies the image and prints the description of the result.

achilleas-k (approved):

cmd/image-builder/describeimg.go
mvo5: I like the idea of the blueprint, that is cool (will require some work in images as currently the partition table + bp is not accessible IIRC).

achilleas-k: Yeah, that's for the future.  Might need a lot of internal work to create a modified image configuration.

supakeen (approved):

PR #184, opened 2025-04-01
GH Action: check the SPEC osbuild/images deps minimum version

thozza:
Add a check which leverages the osbuild/images@check-spec-deps-action action to check that the SPEC files requires at least the minimum versions for dependencies specified by the `osbuild/images`.


achilleas-k:
> Had to bump the required osbuild version -> https://github.com/osbuild/image-builder-cli/actions/runs/14195000453/job/39767992606?pr=184#step:6:221

It's paying off already!

achilleas-k (approved):

PR #186, opened 2025-04-03
main: rename `describe-image` to `describe`

supakeen:
Leaves behind a compatibility alias.

---

Split out from #177.


bcl (approved):
Looks good to me. Maybe add it to the README since we're here?

mvo5 (approved):

PR #189, opened 2025-04-07
many: add support for unpriviledged building (HMS-5955)

mvo5:
This PR adds support for building images without root or a privileged container using the "supermin" binary. It will require access to /dev/kvm though to make the builds reasonably fast.

This is ready for review now, it will need followups to support supermin building in aarch64, s390x and ppc64le, Colin/Johnathan and frieds in COSA did excellent work here so this is hopefully mostly busywork (testing it will be tricky though).

Another followup will be needed to add proper functional tests for supermin, I have a WIP branch in https://github.com/osbuild/image-builder-cli/compare/main...mvo5:supermin-cleanup-test?expand=1 for this but this PR is already quite large so I'm hesitant to make it even larger).

Kudos to cgwalters and jlebon for the excellent help and examples in https://github.com/osbuild/bootc-image-builder/issues/98

/jira-epic COMPOSER-2487

JIRA: [HMS-5955](https://issues.redhat.com/browse/HMS-5955)

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -337,7 +337,6 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	if errors.Is(err, ErrUploadTypeUnsupported) || errors.Is(err, ErrUploadConfigNotProvided) {
 		err = nil
 	}
-
 	if err != nil {
 		return err
 	}
@@ -406,6 +405,23 @@ func cmdDescribeImg(cmd *cobra.Command, args []string) error {
 	}
 
 	return describeImage(res, osStdout)
+}
+
+var osGetuid = os.Getuid
+
+func imageBuilderDefaultCacheDir() (string, error) {
+	if osGetuid() == 0 {
+		return "/var/cache/image-builder/store", nil
+	}
+	xdgCacheHome := os.Getenv("XDG_CACHE_HOME")
+	if xdgCacheHome == "" {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+		xdgCacheHome = filepath.Join(homeDir, ".cache")
+	}
+	return filepath.Join(xdgCacheHome, "image-builder/store"), nil
 }
 
 func run() error {
@@ -483,6 +499,10 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	uploadCmd.Flags().String("aws-region", "", "target region for AWS uploads (only for type=ami)")
 	rootCmd.AddCommand(uploadCmd)
 
+	ibCacheDir, err := imageBuilderDefaultCacheDir()
+	if err != nil {
+		return err
+	}
 	buildCmd := &cobra.Command{
 		Use:          "build <image-type>",
 		Short:        "Build the given image-type, e.g. qcow2 (tip: combine with --distro, --arch)",
@@ -494,7 +514,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	buildCmd.Flags().Bool("with-manifest", false, `export osbuild manifest`)
 	buildCmd.Flags().Bool("with-buildlog", false, `export osbuild buildlog`)
 	// XXX: add --rpmmd cache too and put under /var/cache/image-builder/dnf
-	buildCmd.Flags().String("cache", "/var/cache/image-builder/store", `osbuild directory to cache intermediate build artifacts"`)
+	buildCmd.Flags().String("cache", ibCacheDir, `osbuild directory to cache intermediate build artifacts"`)
 	// XXX: add "--verbose" here, similar to how bib is doing this
 	// (see https://github.com/osbuild/bootc-image-builder/pull/790/commits/5cec7ffd8a526e2ca1e8ada0ea18f927695dfe43)
 	buildCmd.Flags().String("progress", "auto", "type of progress bar to use (e.g. verbose,term)")
--- a/pkg/progress/supermin.go
+++ b/pkg/progress/supermin.go
@@ -2,12 +2,16 @@ package progress
 
 import (
 	"archive/tar"
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"syscall"
 	"time"
+
+	"github.com/google/uuid"
 
 	"github.com/osbuild/image-builder-cli/pkg/util"
 )
@@ -79,6 +83,8 @@ osbuild \
   --cache /store \
   /output/manifest.json
 
+echo "" > /output/%s
+
 # trigger clean shutdown via sysreq
 echo _suo > /proc/sysrq-trigger
 # shutdown is async so we need to sleep here or PID=0 dies
@@ -106,8 +112,8 @@ func addInitTar(superminDir, superminInitScript string) error {
 	return nil
 }
 
-func superminPrepare(prepareDir, export string) error {
-	superminInitScript := fmt.Sprintf(superminInitScriptFmt, export)
+func superminPrepare(prepareDir, export, doneStampName string) error {
+	superminInitScript := fmt.Sprintf(superminInitScriptFmt, export, doneStampName)
 
 	// prepare supermin
 	err := util.RunCmdSync(
@@ -250,9 +256,22 @@ func runOSBuildWithSupermin(pbar ProgressBar, manifest []byte, exports []string,
 	}
 	defer os.RemoveAll(superminTmp)
 
+	// It is non-trivial to exit qemu non-zero if osbuild fails, so
+	// instead we just generate a build-done "stamp" file in output
+	// and if that is missing we know something went wrong and show
+	// the log. It has to be a random stamp to avoid having to deal
+	// with pre-existing build-stamp files.
+	buildUUID, err := uuid.NewRandom()
+	if err != nil {
+		return fmt.Errorf("cannot create build uuid: %w", err)
+	}
+	doneStampName := fmt.Sprintf("done-%s.stamp", buildUUID)
+	doneStampPath := filepath.Join(opts.OutputDir, doneStampName)
+	defer os.Remove(doneStampPath)
+
 	// we need to prepare and then build supermin
 	prepareDir := filepath.Join(superminTmp, "prepare")
-	if err := superminPrepare(prepareDir, exports[0]); err != nil {
+	if err := superminPrepare(prepareDir, exports[0], doneStampName); err != nil {
 		return err
 	}
 	runDir := filepath.Join(superminTmp, "run")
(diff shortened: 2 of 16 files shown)


supakeen:
Small question before I take a look, but how does this interact with the cross-build support via `qemu-user-static`?

cmd/image-builder/main.go
```diff
@@ -361,6 +360,17 @@ func cmdDescribeImg(cmd *cobra.Command, args []string) error {
...
 
+func imageBuilderDefaultCacheDir() (string, error) {
+	if os.Getuid() == 0 {
+		return "/var/cache/image-builder/store", nil
+	}
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(homeDir, ".cache/image-builder/store"), nil
```
supakeen, line 371: Let's use `XDG_CACHE_HOME` for this?

supakeen (commented):
Super cool capability wise.

I do have my concerns that we are growing towards pretty diverging paths in how we execute `osbuild` and that we're (at least partially) duplicating isolation mechanisms (e.g. we're running `bubblewrap` inside a virtual machine inside a container.

We should probably plan on how we want to reconcile those eventually?

pkg/progress/supermin.go
```diff
@@ -0,0 +1,301 @@
...
+		return err
+	}
+	return nil
+}
+
+func superminBuild(prepareDir, buildDir string) error {
+	err := util.RunCmdSync(
+		"supermin",
+		"--build", prepareDir,
+		// XXX: what is the right size?
```
cgwalters, line 142: As it's thinly provisioned (right?) it doesn't matter too much to have it be large.

However, the real fix is to use virtiofs for the rootfs I think, which would obviate much of the role of supermin.

mvo5:
> Small question before I take a look, but how does this interact with the cross-build support via `qemu-user-static`?

We will need qemu-user-static inside the supermin machine and then it should work there, I will look into adding a proper integration test for this too.

bcl:
I'm not reviewing the code, but I have comments. Knowing just what it says in the intro I built this and threw it into my F42 VM and ran into a few things:

* I had osbuild 143 installed and running it resulted in a message - `DNF error occurred: InternalError: dnf-json output was empty`, this was solved by upgrading to osbuild 146. So it looks like it needs a check for the minimal supported osbuild version
* The title says 'unpriviledged building' but skimming the commits it looks like it really means 'inside a container' so I think this needs some documentation added to the README to make it clear how to run it.
* I tried running it with `IMAGE_BUILDER_EXPERIMENTAL=supermin` in the F42 vm (with /dev/kvm) and it did a few things, then failed with `error: expected socket did not appear: files missing after 10s: [/var/tmp/supermin1452852129/vfsd_output.sock /var/tmp/supermin1452852129/vfsd_store.sock]

Another observation is that this seems to be making this project more complex :) I'm not sure that making the tool setup the environment to run the tool in really belongs in the same codebase. Adding dependencies on podman, supermin, kvm setup all seem to be things that are separate from the core job of the project.

mvo5:
> I'm not reviewing the code, but I have comments. Knowing just what it says in the intro I built this and threw it into my F42 VM and ran into a few things:

Thanks for giving it a whirl and for the excellent feedback!
 
>     * I had osbuild 143 installed and running it resulted in a message - `DNF error occurred: InternalError: dnf-json output was empty`, this was solved by upgrading to osbuild 146. So it looks like it needs a check for the minimal supported osbuild version

Yeah, I think @thozza also noticed that ibcli is too liberal about the osbuild version it needs, I will look into this.

>     * The title says 'unpriviledged building' but skimming the commits it looks like it really means 'inside a container' so I think this needs some documentation added to the README to make it clear how to run it.

It is designed to support both containers and running on a regular system (or a VM with nested KVM). If it runs on a regular system it will use supermin when run with "uid != 0". But it sounds like the README (and commit message) needs ot make these points much cleaner.
 
>     * I tried running it with `IMAGE_BUILDER_EXPERIMENTAL=supermin` in the F42 vm (with /dev/kvm) and it did a few things, then failed with `error: expected socket did not appear: files missing after 10s: [/var/tmp/supermin1452852129/vfsd_output.sock /var/tmp/supermin1452852129/vfsd_store.sock]

Thanks! My bad, I think this happens when "virtiofsd" is not installed, I will fix the detection of this and make the error cleaner. 
 
> Another observation is that this seems to be making this project more complex :) I'm not sure that making the tool setup the environment to run the tool in really belongs in the same codebase. Adding dependencies on podman, supermin, kvm setup all seem to be things that are separate from the core job of the project.

That is definitely a valid concern! Its a trade-off (and not that much code) but definitely another layer (pun intended) of complexity. Especially when it runs inside the container it adds two levels of nesting and all the dependencies supermin needs (including a kernel!). We are exploring this because people are hesitant to run privileged containers and the barrier for /dev/kvm seems to be lower.

bcl:
> If it runs on a regular system it will use supermin when run with "uid != 0".

Hm, so I was running it as a normal user in the VM and I had to add the experimental flag. Without it I get:
`error: not enough priviledges: must be root with CAP_SYS_ADMIN`

If I install `virtiofsd` and `qemu-kvm` then supermin will run (still using the experimental flag). So it would make things easier to debug if it checked for required features/packages early on.

With all that it did run successfully!

cmd/image-builder/main.go
```diff
@@ -370,6 +369,23 @@ func cmdDescribeImg(cmd *cobra.Command, args []string) error {
...
+	if xdgCacheHome == "" {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+		xdgCacheHome = filepath.Join(homeDir, ".cache")
+	}
+	return filepath.Join(xdgCacheHome, "image-builder/store"), nil
+}
+
```
supakeen, line 388: Can we split this into a separate PR? I like it but I think we'll have some more discussion in this PR.

pkg/progress/supermin.go
```diff
@@ -0,0 +1,338 @@
...
+
+# We cannot put the host osbuild "store" on our /store dir even with
+# virtiofsd as there will be issues with uid remapping and selinux
+# labels.
+mkdir -p /host/store
+mount -t virtiofs osbuild_store /host/store
+# rsync/cp the caches so that our build is faster
+echo "Populate /store from host"
+mkdir /store
+rsync -a --exclude=./tmp /host/store/ /store
```
supakeen, line 69: This will easily make someone run out of disk space right?

pkg/progress/supermin.go
```diff
@@ -0,0 +1,338 @@
...
+  --output-directory /output \
+  --cache /store \
+  /output/manifest.json
+
+echo "" > /output/%s
+
+# trigger clean shutdown via sysreq
+echo _suo > /proc/sysrq-trigger
+# shutdown is async so we need to sleep here or PID=0 dies
+sleep 999
```
supakeen, line 91: Can it really take over 15 minutes to shut down a virtual machine? Shouldn't we consider that to be an error case and set this to something much lower?

pkg/progress/supermin.go
```diff
@@ -0,0 +1,338 @@
...
+  /output/manifest.json
+# rsync
+rsync -a --exclude=./tmp /store/ /host/store
+
+echo "Running osbuild"
+osbuild \
+  --export %s \
+  --output-directory /output \
+  --cache /store \
+  /output/manifest.json
```
supakeen, line 84: Not entirely happy since we've been doing a bit of effort to centralize how we format calling `osbuild`. But this adds a way of calling it that exists entirely outside of that approach. Don't really know a better way either.

pkg/progress/supermin.go
```diff
@@ -0,0 +1,338 @@
...
+			}
+		}
+		if len(missingPaths) == 0 {
+			return nil
+		}
+		if time.Since(start) >= maxWait {
+			return fmt.Errorf("files missing after %v: %v", maxWait, missingPaths)
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
```
supakeen, line 176: I might prefer to do this with `time.Ticker` and a `context.WithDeadline` instead but since we don't *really* use either anywhere else I won't block on it.

Example:

```go
func waitForFiles(ctx context.Context, paths ...string) error {
	ticker := time.NewTicker(100 * time.Millisecond)
    defer ticker.Stop()

	for {
		select {
		case <-ticker.C:
			// work
		case <-ctx.Done():
			// cancelled due to timeout
		default:
            // wait
		}
	}
}
```

supakeen (commented):
Some comments added.

---

I'm growing a bit concerned (much like @bcl) about the amount of ways we're growing to wrap `osbuild`.

Can we (perhaps *after* this PR) take a step back and see how we actually want to do this? I envision that the different 'strategies' of running `osbuild` belong as wrapper(s/-scripts) installed on the system. Each with their particular dependencies that they need.

We can ship them as subpackages, if we wish and `image-builder` (or whichever part of our stack we use to execute `osbuild`) can query for available wrappers and then pick based on some heuristic.

We can then install `osbuild-via-supermin` in our container (which depends on all the `supermin` stuff), or `osbuild-via-qemu-user-static`, etc.

pkg/progress/supermin.go
supakeen: Yea, also it's not *really* worth the effort since we're not using context(s) anywhere else in our codebase(s). If we were it'd be nice that the contexts inherit everywhere and we can cancel them from all the way up the stack.

Let's skip this then.

cmd/image-builder/main.go
mvo5: Sure - I'm curious about your thoughts here, what extra discussion do you have in min? This change  in itself is only useful if in combination with supermin, otherwise there will never be a user cache with actual files (as only manifest can run as non-root without supermin)

supakeen: Only disk image types require privileges, there are a few that don't.

mvo5: That is interesting. We do require bind mount in osbuild (and maybe more for bwrap) so a user container with --privileged might be enough for non-disk types (like contianer). I have not tested this though.

pkg/progress/supermin.go
mvo5: Yes, one way to avoid this would be something like this:
- create empty store inside VM
- use virtiofsd to only map in store/sources 
This way we won't be able to share anything but the cached files/containers but that might be good enough and avoids the data duplication.

pkg/progress/supermin.go
mvo5: This is a tricky one. This should never fail, magic-sysreq is pretty robust. OTOH adding a print after an e.g. sleep 300 to signal something went wrong could be added.

pkg/progress/supermin.go
mvo5: Yeah, we need a new "osbuild/exec" or something module that is just passed to the various places to have only a single implementation of executing osbuild. Its on my list for a while, maybe now is a good time to do it.

supakeen: That doesn't need to be in this PR, as long as we track (somewhere) that we'll unify it :)

PR #191, opened 2025-04-08
main: `describe` non-hidden

supakeen:
The describe format has become quite useful, we should list it in the help page.

---

@mvo5 mentioned this in another PR, perhaps together with #190 it'd be nice to have this subcommand listed in `help`.


mvo5 (approved):
Yeah, I was initially hesitant but I think its quite nice now and also with the YAMLification it will become more and more useful as (eventually) the output can become the input of a new image-type

mvo5:
(but lets have one or more reviews here just to double check with the team)

achilleas-k (approved):

PR #193, opened 2025-04-08
ci: release into epel-9-next

supakeen:
Instead of directly releasing to epel-9 we should release into epel-9-next and do manual merges/releases when/if RHEL and/or CentOS catch up with our dependencies.


thozza (approved):
LGTM

PR #195, opened 2025-04-08
Add unit test for describe-image

thozza:
Add a unit test for testing that providing only the required command line arguments produces expected output.

### DIFF
--- a/cmd/image-builder/main_test.go
+++ b/cmd/image-builder/main_test.go
@@ -23,6 +23,7 @@ import (
 	main "github.com/osbuild/image-builder-cli/cmd/image-builder"
 	"github.com/osbuild/image-builder-cli/internal/manifesttest"
 	"github.com/osbuild/image-builder-cli/internal/testutil"
+	"github.com/osbuild/images/pkg/arch"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/image-builder/main_test.go
```diff
@@ -707,6 +708,38 @@ type: qcow2
...
+	var fakeStdout bytes.Buffer
+	restore = main.MockOsStdout(&fakeStdout)
+	defer restore()
+
+	err := main.Run()
+	assert.NoError(t, err)
+
+	assert.Contains(t, fakeStdout.String(), `distro: centos-9
+type: qcow2
+arch: x86_64`)
```
mvo5, line 740: I wonder if we could avoid this mocking If we just use `fmt.Sprintf("type: qcow2\narch: %s", arch.Current())` here? Its not a big deal and maybe having the ability is useful in more contexts. But it less code and all that :)

mvo5 (commented):
Thank! This looks similar to https://github.com/osbuild/image-builder-cli/pull/190 but the test should definitely go in, I like that a lot!

supakeen:
Thanks, yea I PR'ed the same functionality this morning in #190 (which was merged) but this actually has a test case that I skipped because we weren't mocking it yet :)

cmd/image-builder/main.go
```diff
@@ -350,7 +351,11 @@ func cmdDescribeImg(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	if archStr == "" {
-		archStr = arch.Current().String()
+		archStr = currentArch().String()
+	}
+	distroStr, err = findDistro(distroStr, "")
```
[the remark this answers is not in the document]
supakeen: I think there is a conflict, at least GitHub tells me there is.

cmd/image-builder/main_test.go
thozza: done

supakeen (approved):

PR #196, opened 2025-04-08
main: enable stric toml checking again

mvo5:
This PR enables strict blueprint toml checking again now that nested fields are also marked decoded correctly (c.f. https://github.com/osbuild/blueprint/pull/12). It also switches the exiting blueprint from json to toml and adds the (previously) problematic disk customizations for good measure (those are tested by the blueprint pr#12 already but lets play this safe as we already have the test).


bcl (approved):
Ack!

PR #199, opened 2025-04-10
Containerfile: add `subscription-manager` to the container

mvo5:
This commit should fix the issue that when run from inside the container ibcli is not able to pick-up subscribed content.

Thanks to lzap for diagnosing this.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -18,7 +18,7 @@ RUN mkdir -p /etc/containers/networks
 # Fast-track osbuild so we don't depend on the "slow" Fedora release process to implement new features in bib
 RUN dnf install -y dnf-plugins-core \
     && dnf copr enable -y @osbuild/osbuild \
-    && dnf install -y libxcrypt-compat wget osbuild osbuild-ostree osbuild-depsolve-dnf osbuild-lvm2 \
+    && dnf install -y libxcrypt-compat wget osbuild osbuild-ostree osbuild-depsolve-dnf osbuild-lvm2 subscription-manager \
     && dnf clean all
 
 COPY --from=builder /build/image-builder /usr/bin/


lzap:
I am testing this, do not merge.

mvo5:
> I am testing this, do not merge.

Thanks! This has indeed more layers, I moved it to draft. In bib (where we have a centos-9 container) we have subscription-manager in the container and then run "dnf check" early to init the dnf plugin that syncs the subscriptions from host into the container. Now here this is a problem because the container is fedora-41 and not rhel :/

Moving to centos-9/ubi9 containers is unfortunately also not trivial (in the general case) becaue when we do rootless builds we use supermin and rely on the container kernel to build the image. But if we would use a centos/ubi kernel we couldn't build btrfs systems as there is no kernel support compiled into the rhel kernel for this (AIUI) so rootless builds for fedora would no longer work (at least for btrfs). Tricky all around.

lzap:
> Moving to centos-9/ubi9 containers is unfortunately also not trivial (in the general case) becaue when we do rootless builds we use supermin and rely on the container kernel to build the image.

Containers use hosts kernel, there is no linux kernel in UBI. Is this because of some userspace API that is not present in UBI?

Anyway, I filed the following PR which correct images library subscription secrets so it will use those from the host:

https://github.com/osbuild/images/pull/1414

Now, according to my extensive testing from yesterday, I can say with confidence that if you run a Fedora container via podman on a subscribed host, the dnf plugin does not work correctly. Therefore running any `dnf` command (e.g. `dnf check`) will not cause creation of `redhat.repos` like on the UBI image. My theory is that this is likely because RHSM is confused about the installed product - it is not RHEL therefore it is unable to match any subscriptions/repositories which must be enabled.

Let's try to merge the mentioned patch and retest, I think it should help.

mvo5:
Thanks, closing based on the last comment.

PR #201, opened 2025-04-10
main: add new `--registrations` options

mvo5:
This new flag allows to add a file with registration data. This is meant to eventually hold all sort of registrations like ansible or satelite but initially only contains the redhat subscription. Currently only JSON is supported.

It looks like:
```json
{
  "redhat": {
    "subscription": {
      "activation_key": "ak_123",
      "organization": "org_123",
      "server_url": "server_url_123",
      "base_url": "base_url_123",
      "insights": true,
      "rhc": true,
      "proxy": "proxy_123"
    }
  }
}
```

This is not part of the blueprint (today) because its more ephemeral than the things we usually put into the blueprint.

This allows us to build images that are immediately registered. It also keeps our options open in the future. If we move to a new blueprint format where we support multiple blueprints and also ephemeral data like this the "registrations" flag just becomes an alias for "--blueprint".

[edit: it seems people like the idea so I dropped the RFC and draft from it]

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -17,6 +18,7 @@ import (
 
 	"github.com/osbuild/image-builder-cli/pkg/progress"
 	"github.com/osbuild/images/pkg/arch"
+	"github.com/osbuild/images/pkg/customizations/subscription"
 	"github.com/osbuild/images/pkg/imagefilter"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/ostree"
@@ -97,6 +99,37 @@ func ostreeImageOptions(cmd *cobra.Command) (*ostree.ImageOptions, error) {
 	}, nil
 }
 
+type registrations struct {
+	Redhat struct {
+		Subscription *subscription.ImageOptions `json:"subscription,omitempty"`
+	} `json:"redhat,omitempty"`
+}
+
+func subscriptionImageOptions(cmd *cobra.Command) (*subscription.ImageOptions, error) {
+	regFilePath, err := cmd.Flags().GetString("registrations")
+	if err != nil {
+		return nil, err
+	}
+	if regFilePath == "" {
+		return nil, nil
+	}
+
+	f, err := os.Open(regFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("cannot open registrations file: %w", err)
+	}
+	defer f.Close()
+
+	// XXX: support yaml eventually
+	var regs registrations
+	dec := json.NewDecoder(f)
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(&regs); err != nil {
+		return nil, fmt.Errorf("cannot parse registrations file: %w", err)
+	}
+	return regs.Redhat.Subscription, nil
+}
+
 type cmdManifestWrapperOptions struct {
 	useBootstrapIfNeeded bool
 }
@@ -160,6 +193,10 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 		}
 		customSeed = &seedFlagVal
 	}
+	subscription, err := subscriptionImageOptions(cmd)
+	if err != nil {
+		return nil, err
+	}
 	// no error check here as this is (deliberately) not defined on
 	// "manifest" (if "images" learn to set the output filename in
 	// manifests we would change this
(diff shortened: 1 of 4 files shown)


cmd/image-builder/main.go
```diff
@@ -97,6 +99,37 @@ func ostreeImageOptions(cmd *cobra.Command) (*ostree.ImageOptions, error) {
 	}, nil
 }
 
+type registrations struct {
+	Redhat struct {
```
mvo5, line 103: We could of course add one more level here if we wanted like
```json
{
 "registrations": {
  "redhat": {...},
}
```
no strong opinion here.

achilleas-k: I suppose.  I'm all for the idea in general, but didn't think about the structure for too long.  We should probably consider everything we might want to add going forward before finalising.

lzap: Absolutely, this is the place where we can put AAP as well.

cmd/image-builder/main.go
```diff
@@ -97,6 +99,37 @@ func ostreeImageOptions(cmd *cobra.Command) (*ostree.ImageOptions, error) {
...
+		return nil, nil
+	}
+
+	f, err := os.Open(regFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("cannot open registrations file: %w", err)
+	}
+	defer f.Close()
+
+	// XXX: support yaml eventually
```
lzap, line 123: Why, why it is JSON first YAML maybe and not the other way around (on a new project)? :-)

cmd/image-builder/main.go
```diff
@@ -430,6 +468,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	manifestCmd.Flags().String("ostree-url", "", `OSTREE url`)
 	manifestCmd.Flags().Bool("use-librepo", true, `use librepo to download packages (disable if you use old versions of osbuild)`)
 	manifestCmd.Flags().Bool("with-sbom", false, `export SPDX SBOM document`)
+	manifestCmd.Flags().String("registrations", "", `filename of a registrations file with e.g. subscription details`)
```
lzap, line 471: I wonder if it makes sense to generalize the 2nd input JSON file into something more generic. What if we decide to add another input, would that be a 3rd input config file? What I am trying to say, if we called this I dont know `--metadata` and in the file we had `subscriptions` key, we could freely expand it with other fields later on. Not sure, just an idea.

cmd/image-builder/main_test.go
```diff
@@ -949,3 +953,81 @@ func TestBasenameFor(t *testing.T) {
...
+	defer restore()
+
+	var fakeStdout bytes.Buffer
+	restore = main.MockOsStdout(&fakeStdout)
+	defer restore()
+
+	err = main.Run()
+	assert.NoError(t, err)
+
+	// XXX: manifesttest really needs to grow more helpers
```
lzap, line 1029: Just saying.... :-)

```
$ ag XXX | wc -l
78
```

lzap:
Looks good, make sure to add info into the README.

cmd/image-builder/manifest.go
```diff
@@ -84,9 +90,10 @@ func generateManifest(dataDir string, extraRepos []string, img *imagefilter.Resu
 		return err
 	}
 	var imgOpts *distro.ImageOptions
-	if opts.Ostree != nil {
+	if opts.Ostree != nil || opts.Subscription != nil {
```
supakeen, line 93: Is it OK to have `opts.Ostree` (or `opts.Subscription`) to be `nil`? This is now an `||` so either one could be.

If this is fine then we could drop the check entirely?

cmd/image-builder/main.go
supakeen: I don't particularly see much value in having an extra level of `registrations`, do we want to put things other than registrations in the registrations file? Ansible Automation Platform could be its own top-level key, or it could be a key under Red Hat.

cmd/image-builder/main.go
mvo5: Mostly consistency with existing tech, current blueprints are json/toml so json seems okay, but given that blueprints++ will be YAML I'm fine moving to yaml here already.

cmd/image-builder/main.go
mvo5: My gut-feeling is that starting more targeted is better (YAGNI and all that). I think if we need more we will either do the stacked blueprints with ephemeral data or move to --metadata, and keep --registrations as a alias for --metadata.

cmd/image-builder/main_test.go
mvo5: *cough* yeah :( sorry. I will open a PR with moving this into images right away

lzap: I did not mean you to force to do anything, sorry. Maybe GH issues instead of `XXX` strings could be better but no biggie.

thozza (approved):
I agree with Simon's suggestion to simplify the creation of `ImageOptions`, but that is not a blocker from my PoV. The change looks good to me.

lzap (approved):

PR #202, opened 2025-04-10
manifest: handle "close()" errors in sbomWriter()

mvo5:
This commit adds error handling for the `f.Close()` errors when we write the SBOM. Errors on close for RW fds are rare but we should handle them so we return the result of `f.Close()` now when returning in sbomWriter(). We still keep the `defer f.Close()` to ensure we do not leak file descriptors when e.g. `io.Copy()` fails. In the "happy" case f is closed without an error and then the defer f.Close() runs and will error with "ErrClosed" but we can ignore that.

An alternative implementaiton might be:
```golang
func sbomWriter(outputDir, filename string, content io.Reader) (err error) {
	...
	f, err := os.Create(p)
	if err != nil {
		return err
	}
	defer func() { err = errors.Join(f.Close(), err) }()
	...
	return nil
}
```
no super strong opinion here.

Thanks to Flo for finding this issues!

[releated to https://github.com/osbuild/bootc-image-builder/pull/891]

### DIFF
--- a/cmd/image-builder/manifest.go
+++ b/cmd/image-builder/manifest.go
@@ -39,11 +39,13 @@ func sbomWriter(outputDir, filename string, content io.Reader) error {
 	if err != nil {
 		return err
 	}
+	// ensure we do not leak FDs if the function returns prematurely
 	defer f.Close()
+
 	if _, err := io.Copy(f, content); err != nil {
 		return err
 	}
-	return nil
+	return f.Close()
 }
 
 func generateManifest(dataDir string, extraRepos []string, img *imagefilter.Result, output io.Writer, depsolveWarningsOutput io.Writer, opts *manifestOptions) error {


schuellerf:
would it be
```
defer func() { err = errors.Join(err, f.Close()) }()
```
to catch all cases?

achilleas-k:
> ```go
> func sbomWriter(outputDir, filename string, content io.Reader) (err error) {
> 	...
> 	f, err := os.Create(p)
> 	if err != nil {
> 		return err
> 	}
> 	defer func() { err = errors.Join(f.Close()) }()
> 	...
> 	return nil
> }
> ```

We would want to return `err` here, no?

schuellerf:
@achilleas-k it is returned as it's the named `err` variable

mvo5:
> would it be
> 
> ```
> defer func() { err = errors.Join(err, f.Close()) }()
> ```
> 
> to catch all cases?
> 
> **EDIT**: in this case it can't happen, so I guess `defer func() { err = errors.Join(f.Close()) }()` would be fine and return an error in all cases.

Thanks and sorry, my bad - `err = errors.Join(err, f.Close())` - this is why I prefer the other version, slightly more straightforward :)

mvo5:
> > ```go
> > func sbomWriter(outputDir, filename string, content io.Reader) (err error) {
> > 	...
> > 	f, err := os.Create(p)
> > 	if err != nil {
> > 		return err
> > 	}
> > 	defer func() { err = errors.Join(f.Close()) }()
> > 	...
> > 	return nil
> > }
> > ```
> 
> We would want to return `err` here, no?

Yeah, sorry, the original err should also be included in the join of course, I wrote this in too much of a hurry. I correct the commit message now.

cmd/image-builder/manifest.go
```diff
@@ -39,11 +39,13 @@ func sbomWriter(outputDir, filename string, content io.Reader) error {
 	if err != nil {
 		return err
 	}
+	// ensure we do not leak FDs if the function returns prematurely
 	defer f.Close()
```
schuellerf, line 43: But this will trigger the linter again, right?
why not go for the slightly more bulky but complete
```
defer func() { err = errors.Join(err, f.Close()) }()
```
otherwise the `// nolint:errcheck` would be missing

schuellerf (changes requested):

cmd/image-builder/manifest.go
mvo5: In ibcli we have no linter complaining yet so I did not add this. Once we add the updated linter this (and more places) will need to change.

We could use the  `errors.Join`. I slightly prefer the other one because:
a) the rules about named return vars in go can be a bit confusing (i.e. what shadows what)
b) the `return f.Close()` at the end is slightly more linear, i.e. its imeediately obvious that the close error is returned here
but its only a slight preference, the extra comment in this version is also kinda annoying.

achilleas-k:
> @achilleas-k it is returned as it's the named `err` variable

the final line is `return nil` though

schuellerf:
@achilleas-k yes, this will then be "error.Join()ed" with the result of a close error
https://go.dev/ref/spec#Defer_statements
or plainly overriden :-D 
the defer is running after the `return nil` if I understood this correctly

achilleas-k:
> the defer is running after the `return nil` if I understood this correctly

It is, but we're returning nil.  `err`, the named return value, is gone by then.

achilleas-k (approved):
Let's get this in as is for now and we can revisit the deferred error catching when we bump the linter.

cmd/image-builder/manifest.go
schuellerf: … this way we are doing a double-close - although the error is ignored - not sure if this is the most elegant solution - but it will probably catch many important cases now

achilleas-k: Yes, in the general (successful) case, there will always be a deferred close which produces an error that is ignored.

PR #214, opened 2025-04-29
refactor: use standard logger instead of logrus

lzap:
I would like to propose moving away from logrus in the image-builder-cli (and standalone CLI tools from composer and images as well). The library is in maintenance mode for quite a while and I believe it does not provide a good structured logging capabilities and API (log/slog is better) while for console logging it is an overkill. It was primarily designed to be structured logging library with some human-formatting features (Printf...).

Now, for services like crc/composer, I am currently working on replacing logrus with log/slog structured logging library which is part of Go. However, structured logging is not a great fit for CLI applications since its primary purpose is to machine-readable output (JSON) and although text formatters are available in the standard library, it is not best reading for humans. Example: time=2025-04-29 11:30 msg=Initializing osbuild

The question tho is what to replace logrus with, and what I would like to propose is a simple one: there is a log package in the standard library that is extremely simple. It basically only provides Print / Prinln / Printf plus Error/Panic counterparts which do the same but exits the program. There are no logging levels, because they do not make sense in the console UNIX apps context - programs should be silent by default and diagnostic output or errors should be printed to stderr. The standard logger in Go is configured to send all output to stderr actually. I like this simplicity. Also, the API is so clean and small that it is super easy to create a utility package that can provide multiple loggers with multiple "levels" and outputs if needed.

The initial version of the PR comes with a new package called `olog` (optional log) which stands for optional logger and it is a proxy to `log.Logger` that can be turned off or on depending on the `--verbose` option. While this could have been a simple variable somewhere in the code, its own package enables seamless and short use: `olog.Printf(...)` that nicely match `log.Printf(...)`.

My plan is to the similar PR in the `images` library with similar `olog` package (a copy is perhaps sufficient this API is stable). This way, the logging can be enabled, disabled or redirected from API users.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -13,7 +14,6 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/osbuild/image-builder-cli/pkg/progress"
@@ -24,6 +24,7 @@ import (
 	"github.com/osbuild/images/pkg/ostree"
 
 	"github.com/osbuild/image-builder-cli/internal/blueprintload"
+	"github.com/osbuild/image-builder-cli/internal/olog"
 )
 
 var (
@@ -409,13 +410,8 @@ func cmdDescribeImg(cmd *cobra.Command, args []string) error {
 }
 
 func run() error {
-	// images generates a lot of noisy logs a bunch of stuff to
-	// Debug/Info that is distracting the user (at least by
-	// default, like what repos being loaded)
-	//
-	// Disable for now until we can filter out the usless log
-	// messages.
-	logrus.SetOutput(io.Discard)
+	// Initialize console logger (stderr, no prefix)
+	log.SetFlags(0)
 
 	rootCmd := &cobra.Command{
 		Use:   "image-builder",
@@ -435,7 +431,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	rootCmd.PersistentFlags().StringArray("extra-repo", nil, `Add an extra repository during build (will *not* be gpg checked and not be part of the final image)`)
 	rootCmd.PersistentFlags().StringArray("force-repo", nil, `Override the base repositories during build (these will not be part of the final image)`)
 	rootCmd.PersistentFlags().String("output-dir", "", `Put output into the specified directory`)
-	rootCmd.PersistentFlags().BoolP("verbose", "v", false, `Switch to verbose mode`)
+	rootCmd.PersistentFlags().BoolP("verbose", "v", false, `Switch to verbose mode (more logging on stderr and verbose progress)`)
 	rootCmd.SetOut(osStdout)
 	rootCmd.SetErr(osStderr)
 
@@ -520,12 +516,21 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 
 	rootCmd.AddCommand(describeImgCmd)
 
+	verbose, err := rootCmd.PersistentFlags().GetBool("verbose")
+	if err != nil {
+		return err
+	}
+	if verbose {
+		olog.SetDefault(log.New(os.Stderr, "", 0))
+		// XXX: add once images has olog support
+		//images_log.SetDefault(log.New(os.Stderr, "", 0))
+	}
+
 	return rootCmd.Execute()
 }
 
 func main() {
 	if err := run(); err != nil {
-		fmt.Fprintf(osStderr, "error: %s\n", err)
-		os.Exit(1)
+		log.Fatalf("error: %s\n", err)
 	}
 }
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"os/exec"
 	"strings"
@@ -14,7 +15,6 @@ import (
 
 	"github.com/cheggaaa/pb/v3"
 	"github.com/mattn/go-isatty"
-	"github.com/sirupsen/logrus"
 
 	"github.com/osbuild/images/pkg/datasizes"
 	"github.com/osbuild/images/pkg/osbuild"
@@ -241,7 +241,7 @@ func (b *terminalProgressBar) Stop() {
 		// I cannot think of how this could happen, i.e. why
 		// closing would not work but lets be conservative -
 		// without a timeout we hang here forever
-		logrus.Warnf("no progress channel shutdown after 1sec")
+		log.Printf("WARNING: no progress channel shutdown after 1sec")
 	}
 	b.shutdownCh = nil
 	// This should never happen but be paranoid, this should
@@ -479,7 +479,7 @@ func runOSBuildWithProgress(pb ProgressBar, manifest []byte, exports []string, o
 		i := 0
 		for p := st.Progress; p != nil; p = p.SubProgress {
 			if err := pb.SetProgress(i, p.Message, p.Done, p.Total); err != nil {
-				logrus.Warnf("cannot set progress: %v", err)
+				log.Printf("WARNING: cannot set progress: %v", err)
 			}
 			i++
 		}
--- a/pkg/setup/setup.go
+++ b/pkg/setup/setup.go
@@ -2,6 +2,7 @@ package setup
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -9,8 +10,6 @@ import (
 	"strings"
 
 	"golang.org/x/sys/unix"
-
-	"github.com/sirupsen/logrus"
 
 	"github.com/osbuild/image-builder-cli/pkg/podmanutil"
 	"github.com/osbuild/image-builder-cli/pkg/util"
@@ -136,7 +135,7 @@ func validateCanRunTargetArch(targetArch string) error {
 		// we could error here but in principle with a working qemu-user
 		// any arch should work so let's just warn. the common case
 		// (arm64/amd64) is covered properly
-		logrus.Warningf("cannot check architecture support for %v: no canary binary found", targetArch)
+		log.Printf("WARNING: cannot check architecture support for %v: no canary binary found", targetArch)
 		return nil
 	}
 	output, err := exec.Command(canaryCmd).CombinedOutput()
--- a/pkg/setup/setup_test.go
+++ b/pkg/setup/setup_test.go
@@ -3,12 +3,12 @@ package setup_test
 import (
 	"bytes"
 	"fmt"
+	"log"
 	"os"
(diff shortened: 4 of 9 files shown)


pkg/setup/setup.go
```diff
@@ -136,7 +135,7 @@ func validateCanRunTargetArch(targetArch string) error {
 		// we could error here but in principle with a working qemu-user
 		// any arch should work so let's just warn. the common case
 		// (arm64/amd64) is covered properly
-		logrus.Warningf("cannot check architecture support for %v: no canary binary found", targetArch)
+		log.Panicf("cannot check architecture support for %v: no canary binary found", targetArch)
```
mvo5, line 138: Please do not change a warning into a panic in the middle of a refactor, if we want to do this, a change like this should be its own commit with a proper rational.

cmd/image-builder/main.go
```diff
@@ -409,13 +411,11 @@ func cmdDescribeImg(cmd *cobra.Command, args []string) error {
...
-	// Debug/Info that is distracting the user (at least by
-	// default, like what repos being loaded)
+	// Dependency library 'images' generates a lot of noisy logs a bunch of stuff to
+	// Debug/Info that is distracting the user (at least by default, like what repos being loaded).
 	//
-	// Disable for now until we can filter out the usless log
-	// messages.
-	logrus.SetOutput(io.Discard)
+	// Forward to the optional logger that can be turned on via options.
+	logrus.SetOutput(olog.Default().Writer())
```
mvo5, line 418: Do we still need this? Could we just remove logrus entirely?

pkg/setup/setup.go
lzap, line 138: Accidentally changed behavior from just a warning to panic, will fix.

pkg/util/util.go
```diff
@@ -17,7 +17,7 @@ func IsMountpoint(path string) bool {
 // Synchronously invoke a command, propagating stdout and stderr
 // to the current process's stdout and stderr
 func RunCmdSync(cmdName string, args ...string) error {
-	logrus.Debugf("Running: %s %s", cmdName, strings.Join(args, " "))
+	olog.Printf("Running: %s %s", cmdName, strings.Join(args, " "))
```
lzap, line 20: This is optional log statement, it will only appear when `--verbose` is provided.

cmd/image-builder/main.go
```diff
@@ -520,12 +520,19 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
...
+	}
+
 	return rootCmd.Execute()
 }
 
 func main() {
 	if err := run(); err != nil {
-		fmt.Fprintf(osStderr, "error: %s\n", err)
-		os.Exit(1)
+		log.Fatalf("error: %s\n", err)
```
mvo5, line 536: When quickly playing with this it seems by default we now output with data/time prefix:
```console
$ go build && ./image-builder xxx
2025/04/29 17:15:01 error: unknown command "xxx" for "image-builder"
```
I would prefer if we would not do this by default as it adds quite a bit of visual noise.

cmd/image-builder/main.go
```diff
@@ -520,12 +520,19 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 
 	rootCmd.AddCommand(describeImgCmd)
 
+	verbose, err := rootCmd.Flags().GetBool("verbose")
```
mvo5, line 523: This will need to be  `Flags`->`PersientFlags`

cmd/image-builder/main.go
```diff
@@ -520,12 +520,19 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
...
+	if err != nil {
+		log.Panicln(err)
```
mvo5, line 525: I personally would just return the err here but of course that is a place where a panic is also apporpriate

internal/log/olog/proxy.go
```diff
@@ -0,0 +1,83 @@
+package olog
+
+import "io"
+
+// SetOutput is a proxy for log.SetOutput.
```
mvo5, line 5: It seems the only parts we use from this currently is "Printf()", can we YAGNI this and just remove everything here (with a comment maybe) and just keep "Prinf()" for now and add as needed?

mvo5 (commented):
Thanks, very excited that we can get rid of logrus! A few inline questions/comments, maybe worthwhile to have a high-bandwidth sync.

cmd/image-builder/main.go
lzap: Sure, this is configurable. Let me do that.

internal/log/olog/proxy.go
lzap: Yeah I wanted to implement all methods because I very often see APIs defining interface for the standard library logger struct in various detail, some only have `Print` functions, others include `Panic` and some actually have all of the methods.

This API will not change anytime soon, I am pretty sure. So we could just leave that there. Or a compromise: just keep all the logging methods and leave setters/getters, except `Writer` that is a useful one.

internal/log/package.go
```diff
@@ -0,0 +1,2 @@
+// Logging utility functions
+package log
```
achilleas-k, line 2: Is the nesting necessary?  Couldn't we have `internal/olog` be the package, instead of `internal/log/olog`?

lzap: Sure thing.

cmd/image-builder/main.go
lzap: I will go ahead today and create a PR in images too, then depending on the order of merges I will update this. Or if it is okay to see some noisy logrus output for a while, I can remove this straight away.

cmd/image-builder/main.go
lzap: Fixed:

```
lzap@dev:~/image-builder-cli$ go run ./cmd/image-builder/ xxx 2>/dev/null

lzap@dev:~/image-builder-cli$ go run ./cmd/image-builder/ xxx
error: unknown command "xxx" for "image-builder"
exit status 1
```

lzap:
So I rebased and moved from `log/olog` to `olog`. The default logger does not have any time/date prefixes now (that was an overlook I hate this too). I cleaned up the new package and broken it down into two files - one is the variable and the other is for the proxy type.

Fixed all the bugs or typos, the initial version should have been draft PR but I was in a rush, this is now much better.

cmd/image-builder/main.go
lzap: Since I already filed a PR for images, I am removing this line and `logrus` completely from this project, adding this commented-out line (can be uncommented once the images PR is merged):

```go
	if verbose {
		olog.SetDefault(log.New(os.Stderr, "", 0))
		// XXX: add once images has olog support
		//images_log.SetDefault(log.New(os.Stderr, "", 0))
	}
```

achilleas-k: Great!

lzap:
There was an alternative proposal to simply just use `log/slog` package and configure it to print just message. A custom handler would be needed for that, but the problem is that this library is strictly for structured logging and the issue is the output would have to be structured. If we only include the `msg` field, then we are loosing a lot, example:

    slog.Info("user created", "username", name)

Would only print

    user created

leaving the additional (and very important) data behind. The only option I see is to format it something like `msg="user created" user="lzap"` or in similar fashion.

Just wanted to drop it here, I think using the simple `log` built-in package is a reasonable approach.

cmd/image-builder/main.go
mvo5: Aha, thanks! I missed that without it we get all the verbosity from "images". I guess we can just make sure images gets merged first and then we update.

pkg/progress/progress.go
```diff
@@ -241,7 +241,7 @@ func (b *terminalProgressBar) Stop() {
 		// I cannot think of how this could happen, i.e. why
 		// closing would not work but lets be conservative -
 		// without a timeout we hang here forever
-		logrus.Warnf("no progress channel shutdown after 1sec")
+		log.Printf("no progress channel shutdown after 1sec")
```
mvo5, line 244: Please add a `WARNING:` prefix here for the message

pkg/progress/progress.go
```diff
@@ -479,7 +479,7 @@ func runOSBuildWithProgress(pb ProgressBar, manifest []byte, exports []string, o
 		i := 0
 		for p := st.Progress; p != nil; p = p.SubProgress {
 			if err := pb.SetProgress(i, p.Message, p.Done, p.Total); err != nil {
-				logrus.Warnf("cannot set progress: %v", err)
+				log.Printf("cannot set progress: %v", err)
```
mvo5, line 482: Same here, please just add a `WARNING:` prefix to the string

pkg/setup/setup.go
```diff
@@ -136,7 +135,7 @@ func validateCanRunTargetArch(targetArch string) error {
 		// we could error here but in principle with a working qemu-user
 		// any arch should work so let's just warn. the common case
 		// (arm64/amd64) is covered properly
-		logrus.Warningf("cannot check architecture support for %v: no canary binary found", targetArch)
+		log.Printf("cannot check architecture support for %v: no canary binary found", targetArch)
```
mvo5, line 138: And here :)

pkg/setup/setup_test.go
```diff
@@ -22,12 +22,14 @@ func TestValidateCanRunTargetArchTrivial(t *testing.T) {
 }
 
 func TestValidateCanRunTargetArchUnsupportedCanary(t *testing.T) {
-	var logbuf bytes.Buffer
-	logrus.SetOutput(&logbuf)
+	logbuf := new(bytes.Buffer)
```
mvo5, line 25: I'm curious (and its not important), any particular reason to go from `var logbuf bytes.Buffer` to `logbuf := new(bytes.Buffer)` ? Is there a technical (or other) advantage or just (personal) preference? No need to change it, I don't mind - just curious if there is something to learn here :) 

I mean, from a purely technical PoV having `var logbuf` on the stack seems "cheaper" because no malloc() call is require and the GC does not need to track it (as its just on the stack and auto cleaned).

lzap: Good catch, no, no reason. Escape analysis is a strong feature we should always take advantage of. Removed here and there (total 2 places).

cmd/image-builder/main.go
```diff
@@ -520,12 +516,21 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
...
 	rootCmd.AddCommand(describeImgCmd)
 
+	verbose, err := rootCmd.PersistentFlags().GetBool("verbose")
+	if err != nil {
+		return err
+	}
+	if verbose {
+		olog.SetDefault(log.New(os.Stderr, "", 0))
+		// XXX: add once images has olog support
+		//images_log.SetDefault(log.New(os.Stderr, "", 0))
```
mvo5, line 526: We can enable this now (as the  images PR is now merged), right? But I guess this can also be a followup

mvo5 (approved):
Thank you

cmd/image-builder/main.go
lzap: Oh great eyes, yes, let me rebase.

lzap: Oh wait we need to upgrade images, do we? Probably a followup, yeah. Let me create a ticket:

https://github.com/osbuild/image-builder-cli/issues/216

PR #218, opened 2025-05-27
image-builder: fix cross-arch uploading

mvo5:
upload: try to auto-detect the upload arch from the filename

This is another convenience feature for the `image-builder upload`
command: when we do not know the target architecture try to
guess it from the filename. This is not perfect but it will
help a lot of users and should be fine until the day we go
and inspect the image.

---

upload: warn the user if no `--arch` flag is passed on upload

This mitigates the issue that a `image-builder upload` command
currently does not know the target architecture of the image
it uploads.

---

upload: default to boot-mode `uefi-preferred` when unset

When using `image-builder upload --to=aws` we do not know
the bootmode. Ideally we would introspect the image for
the boot mode but that is not trivial right now.

So for now just default to platform.BOOT_HYBRID which
translated to `uefi-preferred` in the AWS API calls
which should offer the widest compatbility and should
fix the issue that aarch64 does not boot currently
when uploaded via ibcli.

---

bib: pass the boot-mode to AWS when doing combined
 build/upload

This commit passes the boot mode to AWS when doing a combined
build/upload. Here we know what boot mode to use and we can
pass it easily.

This also adds a "XXX" to think about how to handle what to
do when uploading a pre-existing image where we do not know
the boot mode.

---

image-builder: fix cross-arch uploading

This commit fixes the cross-arch uploading in the most simple
case by reading the `--arch` when `image-builder build --upload`
is used. Note that this is not a complete fix as it will not
take boot mode into account nor will it (by default) DTRT when
`image-builder upload` is used on a previously build images
for a different architecture.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -477,6 +477,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	uploadCmd.Flags().String("aws-ami-name", "", "name for the AMI in AWS (only for type=ami)")
 	uploadCmd.Flags().String("aws-bucket", "", "target S3 bucket name for intermediate storage when creating AMI (only for type=ami)")
 	uploadCmd.Flags().String("aws-region", "", "target region for AWS uploads (only for type=ami)")
+	uploadCmd.Flags().String("arch", "", "upload for the given architecture")
 	rootCmd.AddCommand(uploadCmd)
 
 	buildCmd := &cobra.Command{
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -334,7 +334,8 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	uploader, err := uploaderFor(cmd, res.ImgType.Name())
+	bootMode := res.ImgType.BootMode()
+	uploader, err := uploaderFor(cmd, res.ImgType.Name(), &bootMode)
 	if errors.Is(err, ErrUploadTypeUnsupported) || errors.Is(err, ErrUploadConfigNotProvided) {
 		err = nil
 	}
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -335,7 +335,7 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	}
 
 	bootMode := res.ImgType.BootMode()
-	uploader, err := uploaderFor(cmd, res.ImgType.Name(), &bootMode)
+	uploader, err := uploaderFor(cmd, res.ImgType.Name(), res.ImgType.Arch().Name(), &bootMode)
 	if errors.Is(err, ErrUploadTypeUnsupported) || errors.Is(err, ErrUploadConfigNotProvided) {
 		err = nil
 	}
--- a/cmd/image-builder/upload.go
+++ b/cmd/image-builder/upload.go
@@ -13,6 +13,7 @@ import (
 	"github.com/osbuild/image-builder-cli/pkg/progress"
 	"github.com/osbuild/images/pkg/cloud"
 	"github.com/osbuild/images/pkg/cloud/awscloud"
+	"github.com/osbuild/images/pkg/platform"
 )
 
 // ErrMissingUploadConfig is returned when the upload configuration is missing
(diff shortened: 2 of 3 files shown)


cmd/image-builder/upload.go
```diff
@@ -84,6 +85,16 @@ func uploaderForCmdAWS(cmd *cobra.Command) (cloud.Uploader, error) {
...
 		return nil, err
 	}
+	targetArch, err := cmd.Flags().GetString("arch")
+	if err != nil {
+		return nil, err
+	}
+	if targetArch == "" {
+		// XXX: move this into the awscloud library, i.e.
+		// empty targetArch should default to curernt instead
+		// of leaving it empty
```
supakeen, line 95: This is what `awscloud` already does, I think? See: https://github.com/osbuild/images/blob/main/pkg/cloud/awscloud/uploader.go#L140-L142

mvo5:
> So we (@achilleas-k, @thozza, @kingsleyzissou) and I chatted about this a bit. A big issue here is that we're uploading an 'arbitrary file' without knowing how it was produced thus there's mismatches in the architecture. But even with this PR the boot mode is also set incorrectly on the uploaded AMI (it's not set, which means it defaults to BIOS, which makes them unbootable in most cases).
> 
> I'll add a commit that allows setting the boot mode separately later on as well, however, it seems to me that we should revisit this `upload` command. It seems like it's going to grow an absolutely unwieldy amount of options which are all propagated on the `build` command as well when the other clouds get implemented.
> 
> @mvo5 I think I understand why you called the argument here `--arch`, that's so it uses the same command when `--arch` is passed to `build`; however, would `--arch` apply to all clouds or should it maybe still be prefixed with `--aws-ami-`?

When using `image-builder build --upload` we have the information what image-type is used and can introspect and set the right bios/uefi option in AWS I think. As you point out in the linked issue for the separate upload we loose this information. I wonder if we should include a /.disk directory in our generated images by default with some meta-infomration? Or /var/lib/image-builder,osbuild or somesuch - here we could store this data and introspect it when we upload. Its a (relatively) big change but it seems useful? Having it in the disk image means we need something like guestfstools to easily introspect it but that seems reasonable(?).

supakeen:
> > So we (@achilleas-k, @thozza, @kingsleyzissou) and I chatted about this a bit. A big issue here is that we're uploading an 'arbitrary file' without knowing how it was produced thus there's mismatches in the architecture. But even with this PR the boot mode is also set incorrectly on the uploaded AMI (it's not set, which means it defaults to BIOS, which makes them unbootable in most cases).
> > I'll add a commit that allows setting the boot mode separately later on as well, however, it seems to me that we should revisit this `upload` command. It seems like it's going to grow an absolutely unwieldy amount of options which are all propagated on the `build` command as well when the other clouds get implemented.
> > @mvo5 I think I understand why you called the argument here `--arch`, that's so it uses the same command when `--arch` is passed to `build`; however, would `--arch` apply to all clouds or should it maybe still be prefixed with `--aws-ami-`?
> 
> When using `image-builder build --upload` we have the information what image-type is used and can introspect and set the right bios/uefi option in AWS I think. As you point out in the linked issue for the separate upload we loose this information. I wonder if we should include a /.disk directory in our generated images by default with some meta-infomration? Or /var/lib/image-builder,osbuild or somesuch - here we could store this data and introspect it when we upload. Its a (relatively) big change but it seems useful? Having it in the disk image means we need something like guestfstools to easily introspect it but that seems reasonable(?).

Having a metadata file is a possibility but then requires users to always manage that metadata file and to not lose it; it's still not the nicest thing.

For now this PR seems good; the rest of the discussion about what/how the upload command can handle these cases is probably better suited to the issue.

cmd/image-builder/upload_test.go
```diff
@@ -46,49 +48,79 @@ func (fa *fakeAwsUploader) UploadAndRegister(r io.Reader, status io.Writer) erro
...
+		targetArchArg      string
+		expectedUploadArch string
+		expectedWarning    string
+	}{
+		// simple case: explicit target arch, no warning
+		{"fake-disk.img", "aarch64", "aarch64", ""},
+		// no target arch, detectable from filename: add note
+		{"centos-9-ami-aarch64.img", "", "aarch64", `Note: using architecture "aarch64" based on image filename (use --arch to override)` + "\n"},
+		// no target arch, not detectable form filename: we warn and expect host arch
+		{"fake-disk.img", "", arch.Current().String(), `WARNING: no upload architecture specified, using "x86_64" (use --arch to override)` + "\n"},
```
supakeen, line 65: The error message actually interpolates the current architecture into the string, our tests don't only run on `x86_64` so this fails on every other architure.

cmd/image-builder/main.go
```diff
@@ -477,6 +478,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	uploadCmd.Flags().String("aws-ami-name", "", "name for the AMI in AWS (only for type=ami)")
 	uploadCmd.Flags().String("aws-bucket", "", "target S3 bucket name for intermediate storage when creating AMI (only for type=ami)")
 	uploadCmd.Flags().String("aws-region", "", "target region for AWS uploads (only for type=ami)")
+	uploadCmd.Flags().String("arch", "", `upload for the given architecture`)
```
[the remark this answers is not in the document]
mvo5: Indeed! Fixed

supakeen (approved):

Issue #227, opened 2025-06-20
Development version when using `--version` option should contain more information


thozza:
When using `--version` with the image-builder container, the reported version is `DEVEL`, which is not useful (i.e. when reporting bugs):
```
$ sudo podman run --rm --privileged ghcr.io/osbuild/image-builder-cli:latest --version
image-builder version DEVEL
```

I'd suggest including at least the build commit SHA as the version, potentially with the "next" version that will be used for the release.

supakeen:
I made a small proof of concept locally but there's probably something more to it. What we currently do is set the version through compiler flags (`-X main.BuildVersion=1234`) where `BuildVersion` has a default of `DEVEL`.

In my PoC I have a new `GetVersion(BuildVersion string)` that returns the `BuildVersion` as-is when it's set to anything *but* the sentinel `DEVEL`; otherwise it uses `runtime/debug` to request the build information for the Go binary and takes the `vcs.revision` key as the version. That means you end up with something like: 

```
€ ./image-builder --version      
image-builder version DEVEL-a5987dfa5372866d1d2010ce945404ad6d51e80d
```

When the `vcs.revision` key is not available, or in general `debug.ReadBuildInfo` fails the version is kept at `DEVEL`.

Would this be an OK approach to PR, does it cover all the bases?

I'm also unsure about the sentinel value. I'd like to make it `unknown` so that when things fail it's always unknown, otherwise it's set to some sensical value.

What do you think?

PR #230, opened 2025-06-23
main: `--version` includes git commit

supakeen:
Conditionally include the git commit for the build if no `BuildVersion` was set explicitly.

---

See: #227. This is a proof of concept as I'm not entirely sure what idioms and behaviors I like.

### DIFF
--- a/cmd/image-builder/version.go
+++ b/cmd/image-builder/version.go
@@ -1,0 +1,65 @@
+package main
+
+import (
+	"runtime/debug"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Usually set by whatever is building the binary with a `-x main.version=22`, for example
+// in `make build`.
+var version = "unknown"
+
+type versionDescription struct {
+	ImageBuilder struct {
+		Version      string `yaml:"version"`
+		Commit       string `yaml:"commit"`
+		Dependencies struct {
+			Images string `yaml:"images"`
+		} `yaml:"dependencies"`
+	} `yaml:"image-builder"`
+}
+
+func readVersionFromBinary() *versionDescription {
+	// We'll be getting these values from the build info if they're available, otherwise
+	// they will always be set to unknown. Note that `version` is set globally so it can
+	// be defined by whatever is building this project.
+	commit := "unknown"
+	images := "unknown"
+
+	if bi, ok := debug.ReadBuildInfo(); ok {
+		for _, bs := range bi.Settings {
+			switch bs.Key {
+			case "vcs.revision":
+				commit = bs.Value
+			}
+		}
+
+		for _, dep := range bi.Deps {
+			if dep.Path == "github.com/osbuild/images" {
+				images = dep.Version
+			}
+		}
+	}
+
+	vd := &versionDescription{}
+
+	vd.ImageBuilder.Version = version
+	vd.ImageBuilder.Commit = commit
+
+	vd.ImageBuilder.Dependencies.Images = images
+
+	return vd
+}
+
+func prettyVersion() string {
+	var b strings.Builder
+
+	enc := yaml.NewEncoder(&b)
+	enc.SetIndent(2)
+
+	enc.Encode(readVersionFromBinary())
+
+	return b.String()
+}
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ $(BUILDDIR)/%/:
 
 .PHONY: build
 build: $(BUILDDIR)/bin/  ## build the binary from source
-	go build -ldflags="-X main.BuildVersion=${VERSION}" -o $<image-builder ./cmd/image-builder/
+	go build -ldflags="-X main.version=${VERSION}" -o $<image-builder ./cmd/image-builder/
 
 .PHONY: clean
 clean:  ## Remove all built binaries
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -28,9 +28,8 @@ import (
 )
 
 var (
-	osStdout     io.Writer = os.Stdout
-	osStderr     io.Writer = os.Stderr
-	BuildVersion           = "DEVEL"
+	osStdout io.Writer = os.Stdout
+	osStderr io.Writer = os.Stderr
 )
 
 // basenameFor returns the basename for directory and filenames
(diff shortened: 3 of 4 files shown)


cmd/image-builder/version.go
```diff
@@ -0,0 +1,36 @@
...
+		return v
+	} else {
+		// otherwise try to get it from the build information
+		return versionFromBuildInfo()
+	}
+}
+
+// Try and get a version (commit) from the build info included in the binary since Go 1.18.
+func versionFromBuildInfo() string {
+	version := "DEVEL"
```
mvo5, line 25: We have a very similar function in bib (c.f. https://github.com/osbuild/bootc-image-builder/blob/main/bib/cmd/bootc-image-builder/main.go#L587), having them similar seems nice, I personally think the "buildTime" and "buildTained" are a bit superficial but `build_revision: <rev>` seems nice and keeping it valid yaml is also nice (and we get it almost for free).

(and the test is in https://github.com/osbuild/bootc-image-builder/blob/main/test/test_opts.py#L137)

mvo5 (commented):
Some quick drive-by feedback as I saw this in my notifications :) Hope its helpful

cmd/image-builder/version.go
thozza: FTR, we also have https://github.com/osbuild/osbuild-composer/blob/cedc351bbd5422d04e4df53551fc698d54346ca6/internal/common/constants.go#L30-L38

thozza:
Thank you for working on this! :heart: 

This is a good direction. I'd suggest to:

- Always use git revision in the reported version when building devel RPMs (i.e. in COPR). It makes less sense to include the revision for "stable" RPM builds in the downstream.
- Always use git revision in the reported version when building any devel build, including the container image. Prefixing it with `DEVEL` or alike is IMO OK. If we could also use the version from SPEC and set it at compile time, that would be great (i.e. something like `123.dev.<ref>`
- A very nice thing would be to include also the version of embedded `images` library, just for transparency and easier debugging.
- Having the ability to optionally get the version-related information in machine-readable form would be a great feature.

supakeen:
So, summarizing please correct me if wrong:

1. We want a YAML-alike output for example:

```
version: 1.0
version_images: 1.0
```

2. `version` is *either* whatever was provided as the BuildVersion at build time (e.g. through the spec file) *or* if that's not available then it's the git commit that the binary was built for.

Is that right?

mvo5:
> I've taken a new stab at this, output is like so. Let's discuss. I've tried to merge idioms used across our codebases. Not entirely happy about the global `imagesVersion`; maybe all the work for deps should be done inside `PrettyVersion()` instead?
> 
> ```
> $ make build               
> go build -ldflags="-X main.version=22" -o bin/image-builder ./cmd/image-builder/
> $ ./bin/image-builder --version
> image-builder:
>   version: "22"
>   commit: dc4c0d66c3fdbd9d441c377f5140e0fb61e0f5eb
>   dependencies:
>     images: v0.156.0
> ```

I like this! About the globals, how about just having a `readVersionFromBinary()` helper instead of doing it in init() and then in prettyVersion it just calls readVersionFromBinary() ? This will also make unit testing for prettyVersion trivial and we can have an end-to-end test as part of test/test_container.py probably also almost trivially(?)

cmd/image-builder/version.go
```diff
@@ -0,0 +1,74 @@
...
+	// If we didn't find any dependency info, we set it to unknown
+	if imagesVersion == "" {
+		imagesVersion = "unknown"
+	}
+}
+
+type versionDescription struct {
+	ImageBuilder struct {
+		Version      string `yaml:"version"`
+		Commit       string `yaml:"commit"`
```
mvo5, line 51: maybe omitempty here? I'm not sure we always get this

cmd/image-builder/version.go
```diff
@@ -0,0 +1,74 @@
...
+		Dependencies struct {
+			Images string `yaml:"images"`
+		} `yaml:"dependencies"`
+	} `yaml:"image-builder"`
+}
+
+func PrettyVersion() string {
```
mvo5, line 58: I'm not sure we need this exported(?)

cmd/image-builder/version.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+var (
+	version string
+	commit  string
+
+	imagesVersion string
+)
+
+func init() {
```
mvo5, line 17: Not sure this is needed in init() ? Given that its used in prettyVersion() this could be a normal helper or am I missing something?

cmd/image-builder/version.go
supakeen: It's always set (see `init`) but it will be set to `unknown`. I feel that's probably more useful than omitting? It shows that things weren't build from a git checkout.

cmd/image-builder/version.go
supakeen: We don't! Changed.

cmd/image-builder/version.go
lzap: There was no specific reason, my thinking was if this is a "constant" string, then I would set it in `init` once and never touch it. Returning a copy via a helper is probably safer now when I think about it.

supakeen: I've adjusted `init` to be a helper instead which returns a `*versionDescription` which is then used directly by `prettyVersion`.

mvo5:
Looks super nice now, maybe a tiny test like:
```python
def test_container_version_smoke(build_container):
    output = subprocess.check_output([
        "podman", "run",
        "--privileged",
        build_container,
        "--version",
    ])
    ver_yaml = yaml.load(output, yaml.loader.SafeLoader)
    # XXX: maybe re.match("unknown"|[0-9a-f]{40})
    assert ver_yaml["version"] != ""
    assert ver_yaml["commit"] != ""
    assert ver_yaml["dependencies"]["images"] != "" 
```
in test/test_container.py ?

supakeen:
> Looks super nice now, maybe a tiny test like:
> 
> ```python
> @pytest.mark.skipif(os.getuid() != 0, reason="needs root")
> def test_container_version_smoke(build_container):
>     output = subprocess.check_output([
>         "podman", "run",
>         "--privileged",
>         build_container,
>         "--version",
>     ])
>     ver_yaml = yaml.load(output, yaml.loader.SafeLoader)
>     # XXX: maybe re.match("unknown"|[0-9a-f]{40})
>     assert ver_yaml["version"] != ""
>     assert ver_yaml["commit"] != ""
>     assert ver_yaml["dependencies"]["images"] != "" 
> ```
> 
> in test/test_container.py ?

I've adjusted and included this test, let's keep it as simple as can be for now. This does mean we need pyyaml in the test environment.

mvo5 (approved):
Thank you

PR #239, opened 2025-07-07
Always set rhsm.Facts when generating manifests

thozza:
go.mod: update osbuild/images to v0.156.0

Make sure that images built by ibcli have correct rhsm.Facts set for
distributions that set it (this is handled by the distro definition, not
by the caller of the manifest generator).


achilleas-k (approved):
LGTM

PR #249, opened 2025-07-16
testutil: trivial cleanup/rename

mvo5:
testutil: rename Calls() -> CallArgsList() for clarity

The current MockCmd.Calls() is a bit ambiguous so follow the example
of pytest here and rename to the cleaner CallArgsList() that
captures more closely what the helper is actually doing.

---

testutil: remove unnecessary testutil.MockCmd.Restore()

While looking over the code I noticed that the `Restore()` helper
is not needed (and arguably wrong as it does not reset PATH).
We already use `t.TempDir()` and `t.Setenv()` as part of the
command setup so manually cleanup is not neccessary (and is today
even incomplete). So YAGNI and we can remove Restore().

### DIFF
--- a/cmd/image-builder/main_test.go
+++ b/cmd/image-builder/main_test.go
@@ -392,8 +392,8 @@ centos-9-qcow2-x86_64/centos-9-qcow2-x86_64.qcow2
 `)
 
 	// ensure osbuild was run exactly one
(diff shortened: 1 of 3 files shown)


ondrejbudai (approved):
Thanks!

supakeen (approved):

PR #250, opened 2025-07-16
main: add `--ignore-warnings` cmdline option

mvo5:
This commit adds a new `--ignore-warnings` that can be used to make warnings not an error when `ibcli` runs.


supakeen (approved):
Ignore warnings to me is clear enough that it means that warnings are not errors, or at least I could read it as such.

PR #266, opened 2025-08-01
deps: fix images 0.168

supakeen:
A replacement directive was still present in the `go.mod`, leading to not installing 0.168 but a variation on 0.167.


achilleas-k (approved):
We have a check for this in osbuild-composer I think.  Or maybe it's part of `golangci-lint`?  Either way, let's add something here too.

lzap (approved):

PR #268, opened 2025-08-04
main: add (hidden) pkgsearch command

mvo5:
[draft as this needs tests and also a small change in images]

This commit adds a new hidden pkgsearch command that can be
used to how information about specific packages within the
given distro/arch/imagetype context.

This needs:
https://github.com/osbuild/images/compare/main...mvo5:imagefilter-res-repos?expand=1

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -121,3 +121,5 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -121,3 +121,5 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/osbuild/images => github.com/mvo5/images v0.0.0-20250804145649-0a6396764aa4
```
lzap, line 125: Pretty sure you are aware but just for case...

PR #273, opened 2025-08-07
image-builder: use `github.com/osbuild/blueprint`

mvo5:
Drop using the "images" library blueprint types and use the ones from github.com/osbuild/blueprint/pkg/blueprint instead.

This also moves to the images library to v0.172.0 and the blueprints library to v1.12.0 as this is required for this to work.


achilleas-k (approved):
YAY!

PR #281, opened 2025-08-19
main: fix --extra-repos support when cross building

mvo5:
This commit fixes the issue that --extra-repo are not added correctly to the repo registry when doing cross arch building. The issue is that they get added early at the reporegistry level and because we don't know yet what arch will be used it was initially only added to the host arch. This was fine before we supported cross-arch building.

But now that we do support cross-arch building this is of course an issue and produces subtle bugs. This commit is the "easy" fix, the extra repos are just added to for all architectures. This may not be right but because they are optional its the reponsibility of the user to provide the right ones. Eventually we should probably be smarter here and move extra repo support into manifestgen but that requires changes in images and an images release so this is good enough for now.

Thanks to Simon de Vlieger for reporting this.

### DIFF
--- a/cmd/image-builder/main_test.go
+++ b/cmd/image-builder/main_test.go
@@ -753,7 +753,7 @@ func TestProgressFromCmd(t *testing.T) {
 	}
 }
 
-func TestManifestExtraRepo(t *testing.T) {
+func TestManifestExtraRepos(t *testing.T) {
 	if testing.Short() {
 		t.Skip("manifest generation takes a while")
 	}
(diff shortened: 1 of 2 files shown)


supakeen (approved):
Tested locally with the following reproducer which now selects the correct package:

```
./image-builder manifest \
    --arch aarch64 \
    --distro fedora-42 \
    --use-librepo=false \
    --extra-repo 'https://download.copr.fedorainfracloud.org/results/pbrobinson/a64-kernel/fedora-42-aarch64/' \
    minimal-raw-zst \
    | jq . \
    | rg kernel-core
```

PR #293, opened 2025-09-04
image-builder: use `manifesttest` from `images`

mvo5:
The `manifestest` package got moved to images a while ago, use that intead of our copy.


ondrejbudai (approved):
Deleting code, yay! Thanks!

PR #302, opened 2025-09-16
deps: update to images 0.191.0

supakeen:
This includes the dropping of `shim-ia32` for Fedora 44 (current rawhide).

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -252,7 +252,7 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


cmd/image-builder/main.go
```diff
@@ -252,7 +252,7 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 		if err != nil {
 			return nil, err
 		}
-		img = &imagefilter.Result{distro, archi, imgType}
+		img = &imagefilter.Result{distro, archi, imgType, nil}
```
lzap, line 255: Nitpick: I think at this point it is the right time to change this to multiline form with names where the new field could be simply skipped.

PR #316, opened 2025-09-23
progress: detect real terminal width for messages

lzap:
There is a common pattern in the progress bar code to shorten strings to fit into terminal width. This commit introduces a new utility function for that. It also introduces a fallback terminal width of 80 columns in case the terminal size cannot be determined.

Added a test for the function and made sure it never returns string longer than the specified length. That was not the case for the previous implementation.

Additionally, the progress bar implementation was updated to dynamically detect the terminal width and adjust message lengths accordingly. This ensures that progress messages are fully visible without being truncated, enhancing user experience during long-running operations.

There is few characters left for the spinner and padding, so we subtract 10 from the terminal width when setting message lengths.

A new dependency was introduced, but it was already an indirect one.

Also, turns out the container test does not actually run with any terminal size variables, so I added a fixed terminal size to the podman run command.

---

Split out of: https://github.com/osbuild/image-builder-cli/pull/303

### DIFF
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -10,7 +10,8 @@ import (
 	"time"
 
 	"github.com/cheggaaa/pb/v3"
-	"github.com/mattn/go-isatty"
+
+	"github.com/osbuild/image-builder-cli/pkg/util"
 )
 
 var (
@@ -71,15 +72,16 @@ type ProgressBar interface {
 	Stop()
 }
 
-var isattyIsTerminal = isatty.IsTerminal
-
 // New creates a new progressbar based on the requested type
 func New(typ string) (ProgressBar, error) {
+	updateTerminalSize(os.Stdout.Fd())
+	w, h := getTerminalSize()
+
 	switch typ {
 	case "", "auto":
 		// autoselect based on if we are on an interactive
 		// terminal, use verbose progress for scripts
-		if isattyIsTerminal(os.Stdin.Fd()) {
+		if isattyIsTerminal(os.Stdin.Fd()) && w > 0 && h > 0 {
 			return NewTerminalProgressBar()
 		}
 		return NewVerboseProgressBar()
@@ -144,32 +146,29 @@ func (b *terminalProgressBar) SetProgress(subLevel int, msg string, done int, to
 	return nil
 }
 
-func shorten(msg string) string {
-	msg = strings.Replace(msg, "\n", " ", -1)
-	// XXX: make this smarter
-	if len(msg) > 60 {
-		return msg[:60] + "..."
-	}
-	return msg
-}
-
 func (b *terminalProgressBar) SetPulseMsgf(msg string, args ...interface{}) {
-	b.spinnerPb.Set("spinnerMsg", shorten(fmt.Sprintf(msg, args...)))
+	b.spinnerPb.Set("spinnerMsg", fmt.Sprintf(msg, args...))
 }
 
 func (b *terminalProgressBar) SetMessagef(msg string, args ...interface{}) {
-	b.msgPb.Set("msg", shorten(fmt.Sprintf(msg, args...)))
+	b.msgPb.Set("msg", fmt.Sprintf(msg, args...))
+}
+
+func shortenString(msg string) string {
+	width, _ := getTerminalSize()
+	msg = strings.ReplaceAll(msg, "\n", " ")
+	return util.ShortenString(msg, width)
 }
 
 func (b *terminalProgressBar) render() {
 	var renderedLines int
-	fmt.Fprintf(b.out, "%s%s\n", ERASE_LINE, b.spinnerPb.String())
+	fmt.Fprintf(b.out, "%s%s\n", ERASE_LINE, shortenString(b.spinnerPb.String()))
 	renderedLines++
 	for _, prog := range b.subLevelPbs {
 		fmt.Fprintf(b.out, "%s%s\n", ERASE_LINE, prog.String())
 		renderedLines++
 	}
-	fmt.Fprintf(b.out, "%s%s\n", ERASE_LINE, b.msgPb.String())
+	fmt.Fprintf(b.out, "%s%s\n", ERASE_LINE, shortenString(b.msgPb.String()))
 	renderedLines++
 	fmt.Fprint(b.out, cursorUp(renderedLines))
 }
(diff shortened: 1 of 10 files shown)


bcl (approved):
Seems to work ok running it locally. But I also don't see anything wrong, so I'm mostly saying it didn't obviously break things :)

pkg/util/util.go
```diff
@@ -35,3 +35,15 @@ func OutputErr(err error) error {
 	}
 	return err
 }
+
+// ShortenString shortens a string to the specified length, replacing
+// newlines with spaces. If the string is longer than length, it appends
+// a unicode ellipsis character. If length is 0, it returns the unmodified
+// string.
+func ShortenString(msg string, length int) string {
```
mvo5, line 43: (nitpick) I'm not sure we should move it out, the side-effect of \n->" " for a general util is a bit weird.

If we move it out, maybe into `strutil` to not have a kitchen-sink `util`? I think you mention that you generally preferred it this way in another PR.  And maybe `strutil.ShortenAndNormalize()` or something (this one is a bit long) to indicate that there is this other side-effect

mvo5 (commented):
Thank you! This looks nice, some idea inline for your consideration but I think we could merge as is (but I think its worth thinking about the inline comments)

lzap:
Locally everything works, so not sure what is wrong on the CI/CD. So I am trying out something crazy. I updated the progress integration test to use `asciinema` utility that will record both sessions into a "cast" file. I also updated the GHA so the recordings are archived as artifacts. These are in fact JSON Seq files and they can be played via `asciinema` utility (in Fedora, Ubuntu etc) or pasted (they are relatively small) via https://asciinema.org online tool for viewing.

supakeen:
> Locally everything works, so not sure what is wrong on the CI/CD. So I am trying out something crazy. I updated the progress integration test to use `asciinema` utility that will record both sessions into a "cast" file. I also updated the GHA so the recordings are archived as artifacts. These are in fact JSON Seq files and they can be played via `asciinema` utility (in Fedora, Ubuntu etc) or pasted (they are relatively small) via https://asciinema.org online tool for viewing.

Using asciicinema runs a small risk of it setting up a pty and the error disappearing. It is possible that in CI (and other cases) that there's no controlling pty at all (in normal cases) and maybe no tty on the receiving end either.

lzap:
> What was the exact error you were seeing?

It was this job: https://github.com/osbuild/image-builder-cli/actions/runs/17951169988/job/51050562633

lzap:
> It's not really about `asciinema`. More that it might influence how terminal detection and things work since it might spawn (another) pty.

But that is the goal, podman does not allow too much configuration of tty. Asciinema allows setting the size, other than that it does not change anything.

Here is the recording which I downloaded from the latest (successful) run:

https://asciinema.org/a/DImVCmlxcIsyWSY2iTdwRPQyD

supakeen:
> > It's not really about `asciinema`. More that it might influence how terminal detection and things work since it might spawn (another) pty.
> 
> But that is the goal, podman does not allow too much configuration of tty. Asciinema allows setting the size, other than that it does not change anything.
> 
> Here is the recording which I downloaded from the latest (successful) run:
> 

This code *should* work in environments where there isn't a size (or that don't report a size) right? It can fall back to non-fancy (`verbose`) output in those cases.

pkg/progress/progress.go
```diff
@@ -144,32 +147,37 @@ func (b *terminalProgressBar) SetProgress(subLevel int, msg string, done int, to
...
 }
 
-func shorten(msg string) string {
-	msg = strings.Replace(msg, "\n", " ", -1)
-	// XXX: make this smarter
-	if len(msg) > 60 {
-		return msg[:60] + "..."
+func getTerminalWidth() int {
+	width, _, err := term.GetSize(int(os.Stdout.Fd()))
+	if err != nil {
```
supakeen, line 152: It might be safer to assume that the terminal is in fact broken if it's not responding correctly to `TIOCGWINSZ` and instead downgrade to a non-interactive terminal.

By that I mean that *if* progress is set to auto we currently only check if isatty returns true on the fd; we should probably call GetSize there as well to see if that terminal is behaving correctly.

supakeen:
I put some small comments from my experiences of doing things with terminals, I need to do a deeper look and rack my brain for other issues (I distinctly remember cases where nested pty's are used from another project).

Note that these things are all very 'maybe' and very environment dependent. I'm just slightly worried that suddenly `image-builder` breaks in some environment. I don't think the odds of that happening are *high* since we always pass `--progress=verbose` in Koji and don't really do anything relative to the window sizes we read but only clear a line and write a new line.

It might be good to be a bit defensive and then I'm happy enough to merge.

test/test_container.py
```diff
@@ -76,30 +78,62 @@ def test_container_build_generates_manifest(tmp_path, build_container):
...
+
+    assert os.path.exists(cast_filename)
+    with open(cast_filename, "r", encoding="utf-8") as f:
+        cast_text = f.read()
+
+    assert case.needle in cast_text
+    assert case.forbidden not in cast_text
 
 
-# only test a subset here to avoid overly long runtimes
```
lzap, line 102: And this one accidentally removed, will put back.

pkg/util/util.go
mvo5: Its fine as is, thanks for the tweaks about the `\n`

supakeen:
@lzap Could you fix the conflict here, I'm about ready to ack it but otherwise I have to do it a few times.

test/test_container.py
```diff
@@ -73,26 +75,59 @@ def test_container_build_generates_manifest(tmp_path, build_container):
...
 ])
 @pytest.mark.skipif(os.getuid() != 0, reason="needs root")
-def test_container_with_progress(tmp_path, build_fake_container, progress, needle, forbidden):
+def test_container_with_progress(tmp_path, build_fake_container, case: ProgressTestCase):
     output_dir = tmp_path / "output"
     output_dir.mkdir()
-    output = subprocess.check_output(podman_run + [
-        "-t",
+
+    podman_command = [ podman_run +
```
supakeen, line 98: Needs to be `podman_run + [` instead of `[ podman_run +`.

supakeen:
One comment (introduced by the rebase) and linters need appeasement.

supakeen (approved):

bcl (approved):
Looks good!

PR #317, opened 2025-09-23
many: metrics option for stage durations

lzap:
This PR introduces a new statistics option for the image building stages, allowing users to get more insights into the duration of each stage.

---

Split out from: https://github.com/osbuild/image-builder-cli/pull/303

I need to re-test, got a bunch of conflicts and I am not confident that the code is okay. Feel free to drop comments tho.

### DIFF
--- a/cmd/image-builder/build.go
+++ b/cmd/image-builder/build.go
@@ -17,6 +17,7 @@ type buildOptions struct {
 
 	WriteManifest bool
 	WriteBuildlog bool
+	Metrics       bool
 }
 
 func buildImage(pbar progress.ProgressBar, res *imagefilter.Result, osbuildManifest []byte, opts *buildOptions) (string, error) {
@@ -38,6 +39,7 @@ func buildImage(pbar progress.ProgressBar, res *imagefilter.Result, osbuildManif
 	osbuildOpts := &progress.OSBuildOptions{
 		StoreDir:  opts.StoreDir,
 		OutputDir: opts.OutputDir,
+		Metrics:   opts.Metrics,
 	}
 	if opts.WriteBuildlog {
 		if err := os.MkdirAll(opts.OutputDir, 0755); err != nil {
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -374,6 +374,10 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	withMetrics, err := cmd.Flags().GetBool("with-metrics")
+	if err != nil {
+		return err
+	}
 	// XXX: check env here, i.e. if user is root and osbuild is installed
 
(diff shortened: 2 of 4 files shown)


lzap:
Ready for re-reviews. This is how it looks now, I think the pipeline should probably go after the stage name, but stage name is delivered with comma already so I would need to parse it or something:

```
Message: Finished pipeline xz
Statistics:
	os: Finished module org.osbuild.rpm: 1m8s
	xz: Finished module org.osbuild.xz: 21s
	build: Finished module org.osbuild.rpm: 10s
	os: Finished module org.osbuild.selinux: 6s
	image: Finished module org.osbuild.copy: 3s
```

thozza:
> Ready for re-reviews. This is how it looks now, I think the pipeline should probably go after the stage name, but stage name is delivered with comma already so I would need to parse it or something:

I don't understand your idea, but having the pipeline at the beginning makes sense. Pipelines are encapsulating stages and are the higher-level component, so it makes sense to present the pipeline name before the stage name. TBH, it may make sense to present it in a tree-like view, where all stages are nested under the pipeline, rather than having an unordered list of stages from various pipelines.
> 
> ```
> Message: Finished pipeline xz
> Statistics:
> 	os: Finished module org.osbuild.rpm: 1m8s
> 	xz: Finished module org.osbuild.xz: 21s
> 	build: Finished module org.osbuild.rpm: 10s
> 	os: Finished module org.osbuild.selinux: 6s
> 	image: Finished module org.osbuild.copy: 3s
> ```

An unfortunate thing IMHO is that the stage is called a `module`, but it originates from `osbuild`, so it is not something that the PR can address.

lzap:
> unordered list of stages

It is ordered by the execution time.

Let me know guys how should I continue this. As I stated already, I would like to use this same code in composer and send the data as trace logs so we can have statistics for the hosted image builder service.

thozza:
> > unordered list of stages
> 
> It is ordered by the execution time.

Sorry, I missed that. I think that the core of the problem from my PoV is that all pipelines are mixed. If the sorting based on execution time were to stay, but the stages were grouped by pipeline name, that would be much better. I'm not sure if and how that would affect your next steps for the service.

lzap:
Rebased, renamed the option to `--with-metrics`.

supakeen (approved):

thozza (commented):
Code-wise, the PR looks good to me.

However, the last commit says `many: statistics option for stage durations`, while it adds `--with-metrics` option. Moreover, `metrics` and `statistics` are both used in different packages. It would probably make sense to use a single common name to describe the feature everywhere, unless I'm missing some good reason to use different names in different places.

supakeen (dismissed):

supakeen (approved):
Thanks :)

achilleas-k (approved):

PR #320, opened 2025-09-24
test: check that we get the expected image types

mvo5:
This test ensures that we have a minimal set of expected image types. This protects against the regression we had to fix in https://github.com/osbuild/images/pull/1895


supakeen (commented):
Approving my own code, again (because linters).

supakeen (dismissed):

achilleas-k (dismissed):

supakeen (dismissed):
It's a bit unfortunate that these test cases run after the build tests. I might want to drop those or split the file apart in a follow-up.

Squashed a (hopefully) last bug here and had to rebase on `main` to actually get images 0.195.0 or these tests fail.

mvo5:
> It's a bit unfortunate that these test cases run after the build tests. I might want to drop those or split the file apart in a follow-up.
[..]

Yeah, excellent point, splinting into a slow/fast test is probably a good idea, but this is good for now, thanks again!

achilleas-k:
Merge conflict

thozza (dismissed):
Looks good.

supakeen (approved):

supakeen (approved):

Issue #323, opened 2025-09-25
How to pass root filesystem for `bootc` builds?


supakeen:
When trying to build Fedora bootable containers I get:

```
€ sudo ./image-builder build --bootc-ref=quay.io/fedora/fedora-iot:43 qcow2
WARNING: bootc support is experimental
[\] Manifest generation step
Message: Building manifest for bootc-based-qcow2
error: error setting root filesystem type: root filesystem type is empty
```

`bootc-image-builder` has a `--rootfs` option. What's the equivalent here?

PR #324, opened 2025-09-25
main: add `--bootc-defaultfs` option

mvo5:
Not all bootc containers declare a default filesystem, most noatbly fedora does not. So this commit adds `--bootc-defaultfs` to set it.

It *might* be interesting in the future to consider making this more broad, i.e. to allow `--defaultfs` for package based distros too (with the generic distro we would set it via
`DistroYAMl.DefaultFStype`). But definitely followup material (and requires more discussion if we want it).

Closes: https://github.com/osbuild/image-builder-cli/issues/323

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -189,6 +189,10 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 	if err != nil {
 		return nil, err
 	}
+	bootcDefaultFs, err := cmd.Flags().GetString("bootc-default-fs")
+	if err != nil {
+		return nil, err
+	}
 	var rpmDownloader osbuild.RpmDownloader
 	if useLibrepo {
 		rpmDownloader = osbuild.RpmDownloaderLibrepo
@@ -255,7 +259,7 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 
 	var img *imagefilter.Result
 	if bootcRef != "" {
-		distro, err := bootc.NewBootcDistro(bootcRef, nil)
+		distro, err := bootc.NewBootcDistro(bootcRef, &bootc.DistroOptions{DefaultFs: bootcDefaultFs})
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


cmd/image-builder/main.go
```diff
@@ -513,6 +522,7 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	manifestCmd.Flags().String("ostree-parent", "", `OSTREE parent`)
 	manifestCmd.Flags().String("ostree-url", "", `OSTREE url`)
 	manifestCmd.Flags().String("bootc-ref", "", `bootc container ref`)
+	manifestCmd.Flags().String("bootc-default-fs", "", `default filesystemm to use for the bootc install (e.g. ext4)`)
```
croissanne, line 525: wee typo, filesystemm -> -m

croissanne (commented):
lgtm, one nitpick

thozza:
> It _might_ be interesting in the future to consider making this more broad, i.e. to allow `--defaultfs` for package based distros too (with the generic distro we would set it via
> `DistroYAMl.DefaultFStype`). But definitely followup material (and requires more discussion if we want it).

While allowing setting the default FS type for a distro (i.e Fedora defaulting to `btrfs` and RHEL to `ext4`) makes sense, and it is definitely something that we should add, I'm missing the point of making this a CLI option. Shouldn't the user do this via BP customizations? I understand that we currently do not provide a convenient way to just change the FS type without specifying additional partitioning properties, but that's something to consider.

Moreover, shouldn't the bootc distro definition have a default FS type defined in it, instead of requiring it to be specified on the CLI?

supakeen:
> Moreover, shouldn't the bootc distro definition have a default FS type defined in it, instead of requiring it to be specified on the CLI?

"Should", not all do (Fedora doesn't).

> While allowing setting the default FS type for a distro (i.e Fedora defaulting to ext4 and RHEL to xfs) makes sense, I'm missing the point of making this a CLI option. Shouldn't the user do this via BP customizations?

Maybe. I like a CLI argument.

thozza:
> > Moreover, shouldn't the bootc distro definition have a default FS type defined in it, instead of requiring it to be specified on the CLI?
> 
> "Should", not all do (Fedora doesn't).

I understand that. I meant in our bootc distro definition.

> > While allowing setting the default FS type for a distro (i.e Fedora defaulting to ext4 and RHEL to xfs) makes sense, I'm missing the point of making this a CLI option. Shouldn't the user do this via BP customizations?
> 
> Maybe. I like a CLI argument.

That's fine. I understand that it is more convenient. Nevertheless, it just adds friction and opens the opportunity to achieve the same thing in a multitude of different ways on-prem, in the Service, etc...

thozza:
> [...]
> 
> Thanks, this is a really good question/consideration. I'm honestly a bit torn, what I don't like is the extra friction of a blueprint because its a little bit more complicated than just `--defaultfs` _but_ I agree about the concern about doing the same in many ways and it could also be a great way to introduce people to blueprints (who may not know they exist and that they are super useful).

This makes sense. I don't see a problem with making specific customizations more convenient to use by providing a CLI argument. Still, we should consider whether this should be an addition to a BP customization that does the same thing.
 
> But the root of the issue is that some containers do not set a default fs - so how about we just pick ext4 as the safe choice if the container does not declare one? We could show a warning with a URL that explain how to change it but that would eliminate all friction and still make it easy for our users to discover how to change the fs, wdyt?

Yes, this is the core of my proposal/idea, which may not make sense for bootc world. To have a sane default that gets used when the container does not declare the default FS, instead of forcing the user to provide it explicitly. Providing a way to change the default IMHO makes sense.

thozza (dismissed):
LGTM

croissanne (dismissed):

supakeen (approved):

supakeen (approved):

supakeen (dismissed):

Issue #242, opened 2025-07-09
Allow setting tags when uploading to AWS


FrostyX:
In Fedora Infrastructure, there is an agreement that all resources should have a `FedoraGroup` tag set. It would save all of us a manual step if `image-builder upload --to aws` could set it for us, e.g. `--aws-tag copr`.

PR #300, opened 2025-09-13
Support uploading to libvirt

FrostyX:
See https://github.com/osbuild/images/pull/1858

PR #327, opened 2025-09-25
Allow setting custom tags when uploading to AWS

FrostyX:
Fix https://github.com/osbuild/image-builder-cli/issues/242
See https://github.com/osbuild/images/pull/1903


FrostyX:
I based the PR on top of #300 to make my life easier. Once that PR gets merged, I will rebase so this one has just one commit.

thozza:
@FrostyX You'll need to update the vendored `osbuild/images` to https://github.com/osbuild/images/releases/tag/v0.200.0

thozza:
@FrostyX, you need to bump the minimum required version of `osbuild` in the SPEC file, because the vendored `osbuild/images` library requires it (See the SPEC file check).

thozza (approved):
LGTM

cmd/image-builder/upload.go
```diff
@@ -93,6 +93,21 @@ func uploaderForCmdAWS(cmd *cobra.Command, targetArchStr string, bootMode *platf
...
 		return nil, err
 	}
+	tags, err := cmd.Flags().GetStringArray("aws-tag")
+	if err != nil {
+		return nil, err
+	}
+	var slicedTags [][2]string
+	for _, tag := range tags {
+		parts := strings.SplitN(tag, "=", 2)
+		if len(parts) != 2 {
```
mvo5, line 103: What is the policy of AWS here? Are values wit `=` allowed? Because if they are we could change this to `if len(parts) < 2` and allow `=` in the values as well.

mvo5 (commented):
This is fine but see https://github.com/osbuild/images/pull/1915#issuecomment-3372092578 - I'm not a fan of the `[2]string` to represent key/val. Not a blocker, I don't want to slow you down but I would like to tweak this at some point.

Also a tiny question inline.

supakeen:
> This is fine but see [osbuild/images#1915 (comment)](https://github.com/osbuild/images/pull/1915#issuecomment-3372092578) - I'm not a fan of the `[2]string` to represent key/val. Not a blocker, I don't want to slow you down but I would like to tweak this at some point.
> 
> Also a tiny question inline.

I'm fine with this for now. We can do the struct follow-up ourselves so we don't have @FrostyX juggle too many dependent PRs.

This PR needs some updates to the go deps (probably because we bumped versions in `main`) but otherwise looks fine.

FrostyX:
I submitted a PR for the `AWSTag` struct
https://github.com/osbuild/images/pull/1939

cmd/image-builder/upload.go
FrostyX: I am still testing this part of the PR but then I will rebase :-)

PR #331, opened 2025-09-26
ci: split apart test cases

supakeen:
Splits the tests into separate files along some arbitrary logical boundaries and runs them separately in GitHub CI so we have faster results and we could (theoretically) mark cross-arch as experimental.

---

Draft since I need to figure out CI and it will conflict a bunch of in-flight PRs. I'd like to merge this *after* those land so I don't cause extra work for others.


.pylintrc
```diff
@@ -1,2 +1,5 @@
 [FORMAT]
 max-line-length=120
+
+[MESSAGES CONTROL]
+disable=R0801
```
bcl, line 5: Could you use 'duplicate-code' instead? I had to look up what R0801 was :)

mvo5:
Thanks for working on this! My only minor suggestion would be to combine the rename/name commits into a single `workflow: rename actions` commit as it seems a bit much clutter in the git history for trivial name changes in the same area.

mvo5 (commented):
Shouldn't there be a bunch of removals in this PR as well :) I only see added lines? Also some of the names seems duplicating prefixes, e.g. `test_build_builds...` given that its already in `test_build.py` maybe just `test_builds_image` (there are more like this I just glanced over them so far)

mvo5 (approved):
Thank you

PR #333, opened 2025-09-29
blueprintload: improve error message in Load

ondrejbudai:
When working with ibcli in podman with a blueprint, I got this error message:

Error: statfs /tmp/bp.toml: no such file or directory

I actually have no idea if podman failed because it couldn't find the file on the host, or if image-builder failed because it couldn't find it in the container.

So let's add a bit more context to the error message.


thozza (approved):

mvo5 (approved):
Thank you! I will try to get back to this at some point and write a tiny test (and also look around if there are more "naked errs" that are not tested or bad.

PR #352, opened 2025-10-23
Rework progress

croissanne:
Opened in conjunction with https://github.com/osbuild/images/pull/1966, to see if ppl think this is a good approach.


go.mod
```diff
@@ -141,3 +141,5 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	libvirt.org/go/libvirt v1.11006.0 // indirect
 )
+
+replace github.com/osbuild/images => github.com/croissanne/images v0.0.0-20251023114920-81eefc443b5e
```
lzap: ...

go.sum
```diff
@@ -245,10 +247,6 @@ github.com/opencontainers/selinux v1.12.0 h1:6n5JV4Cf+4y0KNXW48TLj5DwfXpvWlxXplU
 github.com/opencontainers/selinux v1.12.0/go.mod h1:BTPX+bjVbWGXw7ZZWUbdENt8w0htPSrlgOOysQaU62U=
 github.com/osbuild/blueprint v1.16.0 h1:f/kHih+xpeJ1v7wtIfzdHPZTsiXsqKeDQ1+rrue6298=
 github.com/osbuild/blueprint v1.16.0/go.mod h1:HPlJzkEl7q5g8hzaGksUk7ifFAy9QFw9LmzhuFOAVm4=
-github.com/osbuild/images v0.206.0 h1:F9G6dnqrURRUcWE2eWIPQLzHutSCT0OyWMcITK28uCQ=
-github.com/osbuild/images v0.206.0/go.mod h1:iF6bTLzBtyp9l27fexsD5AzwHEn9+bXF5Jr4HHQecmI=
-github.com/osbuild/images v0.209.0 h1:9BRf+N0op1WbQkc+7zVRBZxg4dqS4lty3i2stF3G9lo=
-github.com/osbuild/images v0.209.0/go.mod h1:tZqcrs3eNUA0VPs1h3YCnbnpAskVVfo36CIi2USSfDs=
 github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
 github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
```
lzap: This should not have been in this commit, there is a separate one I guess you accidentally squashed it in.

pkg/progress/command.go
```diff
@@ -48,92 +35,37 @@ func RunOSBuild(pb ProgressBar, manifest []byte, exports []string, opts *OSBuild
...
-	var mw, buildLog io.Writer
-	var writeMu sync.Mutex
-	if opts.BuildLog != nil {
-		mw = newSyncedWriter(&writeMu, io.MultiWriter(&stdio, opts.BuildLog))
-		buildLog = newSyncedWriter(&writeMu, opts.BuildLog)
-	} else {
-		mw = &stdio
+	var mu sync.Mutex
+	buildLog := opts.BuildLog
+	opts.BuildLogMu = &mu
```
lzap: Can someone explain why `newSyncedWriter ` was written in a way it accepts a mutex? I do not understand why a mutex could be shared for this decorator design pattern. It should have been:

```go
type SyncedWriter struct {
	mu sync.Mutex
	w  io.Writer
}

func NewSyncedWriter(w io.Writer) *SyncedWriter {
	return &SyncedWriter{w: w}
}

func (sw *SyncedWriter) Write(p []byte) (n int, err error) {
	sw.mu.Lock()
	defer sw.mu.Unlock()

	return sw.w.Write(p)
}
```

Than we could drop this passing of mutex, it is created on stack either way.

go.mod
croissanne: As this is still draft, this is just to show what it would look like with https://github.com/osbuild/images/pull/1966. Not actually intending to merge this.

pkg/progress/command.go
lzap: Hmmm, right. Well, how about to tee before sync so each writer has its individual mutex that can be shared? Something like:

```go
syncedStdio := newSyncedWriter(&stdio)
syncedBuildLog := newSyncedWriter(opts.BuildLog)
io.MultiWriter(syncedStdio, syncedBuildLog)
```

Then use `syncedBuildLog` to write separately only to the build log independent of stdio. The example assumes `newSyncedWriter` no longer takes a shared mutex.

croissanne: image
maybe helpful

PR #360, opened 2025-10-30
main: fix missing append of `repositories` when using --data-dirs

mvo5:
This commit fixes the missing appending of `repositories` when `--data-dirs` is used. It also improves the tests a bit to ensure this won't happen again.

Thanks to Achilleas for finding this!

This was missing when we did
https://github.com/osbuild/image-builder-cli/pull/128


achilleas-k (approved):
Tiny nitpick.  Fine to handle in a followup.

Thanks!!  LGTM

PR #361, opened 2025-11-05
Drop the need for --privileged for all subcommands except `build`

### LINKED ISSUE #103: Running without `--privileged` returns a mount error

ondrejbudai:
**main: drop the bash entrypoint**

Instead, use the one we share with bib.

---

**main: prepare osbuild env only for the build subcommand**

Prior this commit, we prepared the environment in the main function.
This meant that image-builder needed a lot of privileges even for
operations like describe. However, the environment changes are only
needed for running osbuild, thus it makes sense to run it only for
the build command.

---

Fixes #103

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -15,7 +15,9 @@ jobs:
   build:
     name: "Build & Test"
     runs-on: ubuntu-latest
-    container: registry.fedoraproject.org/fedora:41
+    container:
+      image: registry.fedoraproject.org/fedora:41
+      options: --privileged
     env:
       # workaround for expired cert at source of indirect dependency
       # (go.opencensus.io/trace)


supakeen:
I don't really grok it either; I understand the code but the commit messages are a bit confusing. I can run `image-builder` fine without being root (for `describe`, `manifest`, `list`, etc).

I'll assume this is only applicable for running inside a container (since that's what the code does) and that the commit/title could make it a bit more clear.

supakeen (approved):

PR #362, opened 2025-11-05
many: update all refs to Fedora 43

ondrejbudai:
It's now GA, so let's switch to it.

I actually decided to change all references, including docs. While this adds noise, it makes the project look alive, so I think I actually prefer this approach.


thozza (dismissed):

supakeen:
It seems like updating images to `0.216` breaks (some of) the progress unit tests. Likely only a message difference (additional `\n`)?

PR #365, opened 2025-11-11
ci: add gobump action

lzap:
As it was suggested, we can take advantage of gobump tool which keeps the `go.mod` Go version in check until we can upgrade all the dependencies, namely `images`.


supakeen (approved):
We also want to disable dependabot?

lzap:
> We also want to disable dependabot?

I do not have permissions to do so, can someone confirm it has been disabled? Thanks.

https://github.com/osbuild/image-builder-cli/settings

PR #374, opened 2025-11-19
cmd: make bootc-image-builder a multi-call binary of ibcli (HMS-9808)

mvo5:
[this is now used in https://github.com/osbuild/bootc-image-builder/pull/1157 and green there]

This PR move the bootc-image-builder code into
cmd/image-builder and changes `main()` so that image-builder
becomes a multi-call binary. When image-builder is called with
argv[0] as bootc-image-builder it will use the cobra commandline
and code from bootc-image-builder.

There is still a bit of duplication in the manifest generation
and building between bib_main.go and main.go but that is hard
to avoid if we want the "new" bib to be as close as possible
to the existing one (e.g. we do things like "canChownInPath()"
in bib which we do not need in ibcli).

One important difference between the two is how `anaconda-iso`
image types for bootc handle the repositories. The `bib` personality
stays the same as the orignal, i.e. it will use the container
repositories for depsolving. This require some complicated mTLS
handling that does not map well into ibcli. Here we do not do
that and instead use our own repos for the detected distro. I.e.
if the detected distro is fedora we build the (rpm) installer from
out fedora repos. We can strive to unify this and add (not so nice)
code into ibcli to support this *or* we just warn in ibcli if
bootc with anaconda-iso is used that this behaves differently
from bib and that people should use bib if they need the old
behavior (I would prefer that). I hope then we can phase out
this rpm based bootc anaconda-iso in favor of the container
based bootc-installer type.

This is https://github.com/osbuild/bootc-image-builder/pull/1131
but directly in ibcli instead of doing the bib intermediate step.
The advantage is that we can "freeze" the bib repo and if we
run into issues with the transition have the last known good
bib code. Once we are happy we just delete all go files in the
bib repo and it just becomes a repo that builds the container
for bib.

It has many commits because it imports the full history of bib/cmd/bootc-image-builder.

/jira-epic HMS-8839

JIRA: [HMS-9808](https://issues.redhat.com/browse/HMS-9808)

### DIFF
--- a/cmd/image-builder/bib_main.go
+++ b/cmd/image-builder/bib_main.go
@@ -181,8 +181,8 @@ func bibManifestFromCobraFor(imgref, buildImgref, installerPayloadRef, imgTypeSt
 			},
 		},
 		RpmDownloader: rpmDownloader,
-		Depsolver: func(solver *depsolvednf.Solver, cacheDir string, depsolveWarningsOutput io.Writer, packageSets map[string][]rpmmd.PackageSet, d distro.Distro, arch string) (map[string]depsolvednf.DepsolveResult, error) {
-			depsolveResult, err = manifestgen.DefaultDepsolver(cacheDir, depsolveWarningsOutput, packageSets, d, arch, solver)
+		Depsolve: func(solver *depsolvednf.Solver, cacheDir string, depsolveWarningsOutput io.Writer, packageSets map[string][]rpmmd.PackageSet, d distro.Distro, arch string) (map[string]depsolvednf.DepsolveResult, error) {
+			depsolveResult, err = manifestgen.DefaultDepsolve(solver, cacheDir, depsolveWarningsOutput, packageSets, d, arch)
 			return depsolveResult, err
 		},
 		// this turns (blueprint validation) warnings into


Makefile
```diff
@@ -99,6 +99,7 @@ $(BUILDDIR)/%/:
 .PHONY: build
 build: $(BUILDDIR)/bin/  ## build the binary from source
 	go build -ldflags="-X main.version=${VERSION}" -o $<image-builder ./cmd/image-builder/
+	(cd ./cmd/image-builder && ln -sf image-builder bootc-image-builder)
```
mvo5: @lzap you mean this symlink here in your comment in https://github.com/osbuild/image-builder-cli/pull/374#issuecomment-3569787371 ? If so the idea behind this one was that it would be nice to provide the bib binary everywhere where we have ibcli so that its easier for users to switch - *but* thinking about this again maybe its not such a good ideas as we want people to move to bib and by making it too easy to just stick with bib we defeat this goal. Idk, happy to remove this again and reconsider.

internal/bibimg/imagetypes.go
```diff
@@ -0,0 +1,104 @@
...
+
+type imageType struct {
+	Export string
+	ISO    bool
+	Legacy bool
+}
+
+var supportedImageTypes = map[string]imageType{
+	// XXX: ideally we would look how to consolidate all
+	// knownledge about disk based image types into the images
```
avitova:
```suggestion
	// knowledge about disk based image types into the images
```
very important suggestion

Makefile
lzap: Yes, I meant this link.

Don't we only support bib via its container? Then introducing this link is actually a step back if we distribute it with CLI RPM and we might accidentally acquire more users of bib who were previously not exposed. That is why I thought just building differently named binary would be enough. In fact, if bib is only consumed via podman, then we could just introduce an environment variable because the name of the binary does not matter as it is never directly used it is `docker` or `podman` what is typed in.

mvo5:
> > This branch cannot be rebased due to conflicts
> 
> (though it's not showing any conflicts)

Its strange, rebasing also did not show a conflict - oh well! I rebased and hopefully its happy again.

cmd/image-builder/upload.go
```diff
@@ -31,7 +31,6 @@ var ErrUploadConfigNotProvided = errors.New("missing all upload configuration")
 // ErrUploadTypeUnsupported is returned when the upload type is not supported
 var ErrUploadTypeUnsupported = errors.New("unsupported type")
 
-var awscloudNewUploader = awscloud.NewUploader
```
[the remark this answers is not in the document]
mvo5: Great catch and silly me, I will fix this (probably in a followup, not sure what we should do about the (not)merge conflict in the meantime :( Silly GH!

PR #375, opened 2025-11-19
doc: document `bootc` specifics (HMS-9740)

supakeen:
Write down a stub for the bootc operations that can be performed in `image-builder` so we have a page to build on in the future and something to refer users to now.

### DIFF
--- a/doc/01-usage.md
+++ b/doc/01-usage.md
@@ -157,6 +157,8 @@ When building a bootable container into an image we try to base everything on th
 
 The container(s) used for the various `--bootc-*` arguments must be in the container storage of the user running `image-builder` before the start of the build. This avoids needing to configure `image-builder` with appropriate credentials or access rights for container registries.
 
+The most important argument is `--bootc-ref`, this is a [reference to the container](https://oras.land/docs/concepts/reference/) that contains the filesystem ending up in the image.
+
 ```console
 $ sudo podman pull quay.io/centos-bootc/centos:stream10
 $ sudo image-builder build --bootc-ref quay.io/centos-bootc/centos:stream10 qcow2


doc/01-usage.md
```diff
@@ -149,6 +149,96 @@ error: options validation failed for image type "iot-raw-xz": ostree.url: requir
 $
 ```
 
+### bootc
+
+`image-builder` supports building images from [bootable containers](https://docs.fedoraproject.org/en-US/bootc/getting-started/). Building bootc-based images works differently from `ostree`-based images and package-based images.
+
+When building a bootable container into an image we try to base everything on the container. Thus the distribution that is being built is not known; you cannot use the `--distro` argument in combination with `--bootc-*` arguments as it would do nothing.
+
+The container(s) used for the various `--bootc-*` arguments must be in container storage for the user you're running `image-builder` as before you start the build. This design choice was made due to the pulling of containers often needing additional authentication that `image-builder` cannot present the user with.
```
achilleas-k: Wording is a bit awkward here (first sentence, mostly).  Let's try moving it around a bit:

```suggestion
The container(s) used for the various `--bootc-*` arguments must be in the container storage of the user running `image-builder` before the start of the build. This avoids needing to configure `image-builder` with appropriate credentials or access rights for container registries.
```

achilleas-k (dismissed):
Neat, thanks!  Small rephrasing suggestion in-line.  Up to you if you prefer it or not :)

lzap (changes requested):
I am not sure about the ref arguments.

Could you also review other arguments of CLI and explicitly state what is supported and what is not? I am specifically interested in registration which I am going to test right now and report back in the PR later today.

doc/01-usage.md
```diff
@@ -149,6 +149,96 @@ error: options validation failed for image type "iot-raw-xz": ostree.url: requir
...
+# ...
+```
+
+#### `bootc-build-ref`
+
+By default `image-builder` uses the container passed in `--bootc-ref` as the buildroot for the build. If your container does not contain the necessary tooling to turn it into other artifacts then you can explicitly pick a container to use as a buildroot with `--bootc-build-ref`.
+
+#### `bootc-installer-payload-ref`
+
+When `image-builder` builds an installer ISO there are two inputs. One is the installer (usually Anaconda) environment (passed as `--bootc-ref`) and the other is the bootable container that will be installed by the installer (passed as `--bootc-installer-payload-ref`). The `--bootc-installer-payload-ref` argument is mandatory when building an installer ISO and should refer to a bootable container that contains the appropriate packages and configuration for an installer.
```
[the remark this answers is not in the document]
supakeen: Re-reading this the sentence you changed was correct, but the one following one was not. I've dropped that second sentence :)

lzap:
Just for the record, I have built a bootc image with registration and it does not work, system is not registered and activation key is ignored when I try to initiate registration.

lzap (dismissed):
Wow am I the only one confused on how `bootc-installer-payload-ref` vs `bootc-ref` work together? The way I read the former is "bootable container with installer as the payload".

What I find confusing is how semantics of `bootc-ref` changes dramatically with the context. When used alone, it is the bootable container referenced that is supposed to be installed, however, when used with anaconda it is actually the reference to anaconda itself.

doc/01-usage.md
```diff
@@ -149,6 +149,103 @@ error: options validation failed for image type "iot-raw-xz": ostree.url: requir
 $
 ```
 
+### bootc
+
+`image-builder` supports building images from [bootable containers](https://docs.fedoraproject.org/en-US/bootc/getting-started/). Building bootc-based images works differently from `ostree`-based images and package-based images.
+
+When building a bootable container into an image we try to base everything on the container. Thus the distribution that is being built is not known; you cannot use the `--distro` argument in combination with `--bootc-*` arguments as it would do nothing.
```
bcl: I'd drop `try to` since that implies that there may be cases where it isn't based on the container.

doc/01-usage.md
```diff
@@ -149,6 +149,103 @@ error: options validation failed for image type "iot-raw-xz": ostree.url: requir
...
+
+The container(s) used for the various `--bootc-*` arguments must be in the container storage of the user running `image-builder` before the start of the build. This avoids needing to configure `image-builder` with appropriate credentials or access rights for container registries.
+
```
bcl: This could use a mention of `--bootc-ref` before actually using it in the example. Something like "--bootc-ref is a reference to the container to be installed in the image"

doc/01-usage.md
supakeen: There are things that aren't based on the container; for example customizations and we do use (some) host tooling such as `skopeo` I believe.

doc/01-usage.md
supakeen: Makes sense, I've added a commit that explains it before the example and links to a page explaining what a container reference is.

PR #391, opened 2025-12-01
test: add test for subscribed content

mvo5:
This is a followup for https://github.com/osbuild/image-builder-cli/pull/390 - it adds a test that ensures that ibcli can access subscribed content in its container from a subscribed system.

This needs a SUBSCRIPTION_{ORG,ACTIVATIONKEY} secret and because of the way that secrets are handled in forks the test will not be run on forks but it will be run in the merge queue.

Note that just running `image-builder manifest` is not enough,
even when this works the librepo/curl download can still fail
if osbuild does not find the right secrets.

Thanks to lzap for reporting this.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -40,6 +40,35 @@ jobs:
         run: |
           # use "-s" for now for easier debugging
           sudo pytest -s -v test/test_manifest.py
+
+  # Run a separate test with subscription-manager so that we can
+  # ensure that rhel can be build from inside a container, its
+  # a bit tricky because podman-in-podman is fiddly and also because
+  # we need to test downloading the rpms, a manifest build is not enough
+  manifest-in-osbuild-container:
+    name: "RHEL tar build in container"
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yaml
```diff
@@ -40,6 +40,32 @@ jobs:
         run: |
           # use "-s" for now for easier debugging
           sudo pytest -s -v test/test_manifest.py
+
+  # those tests give us testing of the subscription-manager too
+  manifest-in-osbuild-container:
+    name: "RHEL manifest in container"
```
supakeen: I guess this now a RHEL build :)

mvo5:
> Why this is working tho? Isn't the test supposed to fail?

See the text in the PR description:
```
This needs a SUBSCRIPTION_{ORG,ACTIVATIONKEY} secret and because of the way that secrets are handled in forks the test will not be run on forks but it will be run in the merge queue.
```

mvo5:
> Ah so it cannot be merged until it is fixed I assume, thanks.

I will put it into the merge-queue, it should get kicked out again because osbuild is not ready. But if its not kicked out something else is wrong :( (or there is a miracle cure ;)

PR #401, opened 2025-12-11
main: error when image-builder is used to create anaconda-iso

mvo5:
The behavior of anaconda-iso without specil mTLS setup is different from bib so instead of introducing subtle incompatibilities just error.

Note that we could support this with something like: https://github.com/osbuild/image-builder-cli/compare/main...mvo5:consolidate-bib-anaconda-iso-wip?expand=1 but the general code that is currently fairly nice becomes not so nice anymore as we need to add a bunch of extra things like extracting mTLS, cleanup etc for an image-type we consider legacy and want to replace with the `bootc-installer` image type.


bcl (approved):
Looks good!

PR #403, opened 2025-12-11
test: add rhel specific tests

mvo5:
This PR adds an regression/integration test for the bootc anaconda-iso image type for #404. As a by-product it also tweaks the "manifest-is-container" test for rhel that was not quite working before (see the git commit message for details).

Please review extra carefully as this is a pull_request_target PR

PR #404, opened 2025-12-11
bib: fix anaconda-iso mTLS key extraction

mvo5:
There was a (subtle) bug in the ibcli version of the mTLS key exaction. It was happening too late, i.e. when the bootc container was already unmounted.

This commit moves the extraction into the `Depsolve` function which is run while the container is mounted which means we can extract the mTLS config.

Note that this was not discovered earlier because we lack and end-to-end test for RHEL based bootc images :(

The test is in https://github.com/osbuild/image-builder-cli/pull/403 - I ran it locally and it works but I wanted to get this fix in ASAP as it blocks https://github.com/osbuild/bootc-image-builder/pull/1157


ondrejbudai (approved):
Thanks!

PR #408, opened 2025-12-15
many: `force-repo-dir` instead of `force-data-dir`

supakeen:
Previously I deprecated `data-dir` and renamed it to `force-data-dir` to have consistent naming for options that overwrite internal state instead of appending to it (arguments that start with `force-` are overwrites, arguments that start with `extra-` are merges).

With `force-data-dir` we require a `repositories` sub directory. After some discussion we came to the idea to instead of enforcing a directory layout on the user we want to be able to point directly at a directory of repositories.

This commit deprecates `force-data-dir` and introduces `force-repo-dir`. Both write to the same internal variable.

We also allow repository files to live *either* in a `repositories` subdirectory to keep backwards compatibility for any of the arguments or to live directly at the top level of the passed directory.

We show a deprecation warning if a `repositories` sub directory exists and inform users that the `repositories` subdirectory will not be wanted in the future and how to resolve the emitted warning.

This option will be used together with the future introduction of `force-defs-dir` to handle definitions (and its `extra-defs-dir` counter part) unifying all options into a consistent naming scheme.

### DIFF
--- a/cmd/image-builder/repos.go
+++ b/cmd/image-builder/repos.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/fs"
 	"net/url"
+	"os"
 	"path/filepath"
 
 	"github.com/osbuild/images/data/repositories"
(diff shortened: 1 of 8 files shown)


cmd/image-builder/repos_test.go
```diff
@@ -71,8 +71,8 @@ func TestNewRepoRegistryImplDatadir(t *testing.T) {
 
 	// create a custom datadir with testdistro-1.json, the basefilename
```
thozza: Nitpick: the comment should be amended, since it still mentions `datadir`.

cmd/image-builder/repos_test.go
```diff
@@ -71,8 +71,8 @@ func TestNewRepoRegistryImplDatadir(t *testing.T) {
...
 	// must match a distro nameVer
-	dataDir := t.TempDir()
-	repoFile := filepath.Join(dataDir, "repositories", "testdistro-1.json")
+	repoDir := t.TempDir()
+	repoFile := filepath.Join(repoDir, "repositories", "testdistro-1.json")
```
thozza: It may be beneficial to test this also without the `repositories` sub-directory.

cmd/image-builder/repos.go
```diff
@@ -53,12 +54,18 @@ func parseRepoURLs(repoURLs []string, what string) ([]rpmmd.RepoConfig, error) {
...
 	var builtins []fs.FS
 
-	if dataDir != "" {
-		repoDirs = []string{filepath.Join(dataDir, "repositories")}
+	if repoDir != "" {
+		withRepoSubdir := filepath.Join(repoDir, "repositories")
+		if _, err := os.Stat(withRepoSubdir); err == nil {
+			// we don't care about the error case here, we just want to know
+			// if it exists; not if we can't read it or other errors
+			fmt.Fprintf(os.Stderr, "WARNING: `force-repo-dir` found a `repositories` subdirectory at '%s', in the future `image-builder` will not descend into this subdirectory to look for repository files. Please move any repository files directly into the directory '%s' and remove the `repositories` subdirectory to silence this warning.\n", withRepoSubdir, repoDir)
```
thozza: Mentioning the `force-repo-dir` option here may be misleading, since one could specify `--data-dir` and get the warning about an option that they didn't use.

supakeen:
> Very long warning, three lines on 80x25 terminal, this will be quite effective. :D
> 
> I am wondering why you used backtics and single quotes is that some kind of standard or convention in CLI? I would probably use `%q` for both to save keystrokes.

I think it's partial markdown and partial representation of strings (in Python).

PR #414, opened 2025-12-17
[RFC] main: print pretty json manifest

mvo5:
This commit makes the json pretty when `image-builder manifest` is used. Given that this is most likley something that our users run to actually see the manifest we should make it pretty by default.

### DIFF
--- a/cmd/image-builder/main_test.go
+++ b/cmd/image-builder/main_test.go
@@ -29,6 +29,15 @@ import (
 	"github.com/osbuild/images/pkg/arch"
 )
 
+func assertJsonContains(t *testing.T, jsonInput, searchTerm string) {
+	t.Helper()
+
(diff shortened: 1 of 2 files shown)


lzap:
Love it but here is an idea. Interactive terminal? Pretty print. Flat otherwise.

lzap (approved):
I mean this is better than before sure, no need to make it too complex. I would go for two spaces myself but you are the inventor here :-)

lzap:
> Haha - I absolutely do not mind, I just copied the json.Ident() from the upstream docs, if you prefer 2 instead of 4 I will gladly change it, just say the word!

As you already have figured out, I am very, very rarely making someone else change their code. I only offer other view and if they do not fall in love in these (often crazy) ideas, that is fine :-)

Let's do it. Thanks.

Issue #413, opened 2025-12-16
Cross-distribution builds broken in some cases due to post-quantum cryptographic keys in combination with certain versions of RPM and cryptographic backends


supakeen:
We've discovered an issue where `image-builder` is unable to build certain RHEL 10.1 image types on Fedora 43, RHEL 9.x, or CentOS Stream 9 systems. While not 'officially' supported to build cross-distribution it is still something we like to have working.

---

With the latest updates in RHEL 10.1 certain packages now contain signatures that are signed with the `ML-DSA-87+Ed448/D246D6276AFEDF8F(2025-10-08) Red Hat, Inc. (release key 4)` which can be found at: https://access.redhat.com/security/team/key this key is currently not included in the built-in repositories shipped with `image-builder`. To handle this we will be including the new key: https://github.com/osbuild/images/pull/2097.

Aside from this we discovered that the default crypto-policies configuration (specifically for sequoia) would block verification of RPMs signed with this key even if the key is present in the built-in repositories. To remediate this I opened https://github.com/osbuild/osbuild/pull/2288 which will be backported to RHEL and make it into Fedora.

All of this means that currently `image-builder` cannot build RHEL 10.1 on any host system but a future release can.

When these changes (updated `osbuild`, updated repository definitions) have landed it will be possible to build RHEL 10.1 on RHEL 9 and RHEL 10. However, building RHEL 9.7 or RHEL 10.1 on Fedora 43 will not yet be possible.

When `image-builder` builds artifacts it tries to do so with as few host dependencies as possible. However, `image-builder` *does* set up an initial minimal "buildroot" using executables from the host. When packages in this buildroot are signed with the aforementioned key the hosts `rpm` and `rpm`'s cryptographic backend must be capable of importing and verifying this key to use it for signature verification.

On Fedora 43 RPM is not able to import the above key, we've considered *not* importing the above key on Fedora systems however RPM does not skip trying to verify the signature: https://issues.redhat.com/browse/RHEL-112730

---

On RHEL 9.x and CentOS 9.x RPM is also unable to import the key; thus we won't use this key to verify packages. In this case RPM correctly ignores the signature it cannot validate and thus these distributions do not suffer from the same issue as Fedora.

---

[...]

PR #415, opened 2025-12-17
[RFC] main: always use a bootstrap buildroot (but support  disabling)

mvo5:
We recently had two issues where creating our buildroot was
broken in certain situations because the local rpm would
not be compatible with the requirements of the package
that got installed in the buildroot:
https://issues.redhat.com/browse/RHEL-128741
https://github.com/osbuild/image-builder-cli/issues/413

The issue here is that the local rpm is something we do
not control but we need a way to "bootstrap" our buildroot
(once we have a buildroot we use that for everything else).

This commit enables the "bootstrap" container feature of
the images library by default to avoid this dependency.
This means that we "podman pull" a minimal container (e.g.
ubi for rhel) that contains python3 and rpm to then use it
to install our real buildroot.

Note that this enables it all the time even if the host
distribution and target distribution match. The reason
is rebuildability - ie. the same manifest should always
produce the same result and in the general case we do
not know if e.g. a manifest that was part of
`image-builder build --with-manifest` is used somewhere
else again.

For restricted environment where pulling a container is a
problem or for situations where its known that the host
rpm is fine we provide: `--without-bootstrap-container`
to disable this function.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -135,17 +135,10 @@ func subscriptionImageOptions(cmd *cobra.Command) (*subscription.ImageOptions, e
 	return regs.Redhat.Subscription, nil
 }
 
-type cmdManifestWrapperOptions struct {
-	useBootstrapIfNeeded bool
-}
-
 // used in tests
 var manifestgenDepsolver manifestgen.DepsolveFunc
 
-func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []string, w io.Writer, wd io.Writer, wrapperOpts *cmdManifestWrapperOptions) (*imagefilter.Result, error) {
-	if wrapperOpts == nil {
-		wrapperOpts = &cmdManifestWrapperOptions{}
-	}
+func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []string, w io.Writer, wd io.Writer) (*imagefilter.Result, error) {
 	dataDir, err := cmd.Flags().GetString("force-data-dir")
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


supakeen (changes requested):
This cannot be enabled by default in locations where we do not have network access such as when running in Koji or during RPM builds. We need a plan for this so I'm requesting changes to block this PR until we have one :)

We will need to update the `koji-image-builder` plugin to pass the `--without-bootstrap-container` argument as we don't have access there. This will then need to be coordinated with Fedora and CentOS release engineering to be deployed before we can publish a version of `image-builder` with this behavior.

This is a chicken-and-egg problem; we cannot unconditionally set this argument until it exists and I'd *really* prefer to not do any version inspection in `koji-image-builder` so I hope someone has a cool idea around this :)

---

As an aside, if we're doing this then our bootstrap containers *must* be under the control of the distribution we're building with them. Many of them currently are not leading to 3rd party containers with non-direct relationships being used? That irks me a bit.

We also open our selves up to not having direct control over these bootstrap containers so if anything we use from them (we've previously had to request inclusion of certain packages in them) disappears builds break. To *me* this is a more likely scenario since it has happened a few times already then the current case where RPM is broken in all distributions except RHEL.

While it is probably generally an OK idea to always bootstrap as it squats a class of bugs, it also means we won't find certain bugs.

Overall our approach has worked OK over the years and this change is probably (?) mostly motivated by two (quite recent) incidents, both related to the same RHEL change (PQC) where I personally might not see a pattern yet.

Conan-Kudo:
The way we do it in kiwi is we use kiwi itself to create the bootstrap environment. Then use the bootstrap environment to create the "real" one. I'd recommend that for osbuild too.

PR #416, opened 2025-12-17
main: unif/refactor std{out,err} test handling

mvo5:
[draft as this is build on top of https://github.com/osbuild/image-builder-cli/pull/415 and  there would be conflicts otherwise, if 415 does not go in I think its still worth doing this (it just requires to remove the one extra test added in 415]

For historic reasons we have two way to test the std{out,err}
output. One is to just monkey patch `var osStdout = os.Stdout`
and then replace it in the tests, the other to have a helper
that works like a `with` in python that replaces the stdout/stderr
dynamically, e.g.:
```go
stdout, stderr := testutil.CaptureStdio(t, func() {
        err = main.Run()
        require.NoError(t, err)
})
```
The trouble is that they are not quite compatible. In places
where we use the monkey patches `osStdout` the
testutil.CaptureStdio() will not work as it (temporarely)
replaces the "real" `os.Stdout` (and vice-versa).

So this commit unifies all tests to use testutil.CaptureStdio()
which feels slightly nicer than the other approach in the
sense that one does not need to remember to use `osStdout` in
the code. Its mostly mechanical.

### DIFF
--- a/cmd/image-builder/bib_cloud.go
+++ b/cmd/image-builder/bib_cloud.go
@@ -42,11 +42,11 @@ func bibUpload(uploader cloud.Uploader, path string, flags *pflag.FlagSet) error
 		size = st.Size()
 		pbar.SetTotal(size)
 		pbar.Set(pb.Bytes, true)
-		pbar.SetWriter(osStdout)
+		pbar.SetWriter(os.Stdout)
 		r = pbar.NewProxyReader(file)
 		pbar.Start()
 		defer pbar.Finish()
 	}
 
-	return uploader.UploadAndRegister(r, uint64(size), osStderr)
+	return uploader.UploadAndRegister(r, uint64(size), os.Stderr)
 }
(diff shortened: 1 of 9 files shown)


mvo5:
> I mean it is better, but I really do not like mocking global functions like that. These tests should have not been unit tests but fully integration tests actually running the binary via exec syscall. This way, we would not need to worry about this, every test would have been a new process.

When this was written it was a bit unclear if we want to embrace pytest based integration tests. It seems we have landed on this so most of these could just be written in the style of the current `test/test_smoke.py` (or similar). 

> But still that is a workaround to the main problem we have in the design here: any functionality that does print anything to stdout/stder should have a parameter with io.Writer so we can test anything we want.

Well, its more of a trade-off than a "problem". When testing a cli tool that always prints to stdout/stderr, when adding an extra writer (well two actually for both stdout/stderr) makes the common case (stuff is not run under testing) look less nice, hence this optimizes for this (on the expense of the test setup that you dislike). But of course one can decide a different trade-off if the values/weights are different.

lzap:
Pulled the commit in question out of this PR into: https://github.com/osbuild/image-builder-cli/pull/450

PR #417, opened 2025-12-17
main: embed manifestgen.options into manifestOptions

mvo5:
[this is part of https://github.com/osbuild/image-builder-cli/pull/415 but given that this one is controversial (as expected) I open this as a separate PR because I think its nice on its own]

When we generate the osbuild manifest we need to take a bunch of commandline options into account. These are collected and passed via `manifestOptions` in the CLI. There is a (small) overlap with options that are then "converted" to options that need to be passed to `manifestgen` via `manifestgen.Options`.

Previously there was manual code for this but its slightly nicer to just embedd the `manifestgen.Options` into the more general `manifestOptions` of the CLI. This avoid some boilerplate code and might serve as a useful pattern in other places. So this commit does that now (even though the wins are not that big).

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/osbuild/images/pkg/customizations/subscription"
 	"github.com/osbuild/images/pkg/distro/bootc"
 	"github.com/osbuild/images/pkg/imagefilter"
+	"github.com/osbuild/images/pkg/manifestgen"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/ostree"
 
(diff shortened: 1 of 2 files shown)


cmd/image-builder/main.go
```diff
@@ -304,27 +308,31 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 	}
 
 	opts := &manifestOptions{
+		ManifestgenOptions: manifestgen.Options{
```
lzap: Frankly, I do not like Options suffix. We have one extreme when everything is `SomethingCustomizations`. I suggest to consider naming this just `Manifestgen`

PR #418, opened 2025-12-17
data: install tmpfiles.d/image-builder.conf to auto-clean cache

mvo5:
Our osbuild/image-builder cache is located in /var/cache/image-builder and it can grow quite big. To avoid it growing out of bounds too much add a tmpfile.d snippet to ensure that we clean content older than 14d. The 14d is a bit arbitrary, happy to change to longer or shorter cache timeouts (side-note: we have a cache limit in osbuild and an LRU there too but sources are currently not part of the cache limits in osbuild).

### DIFF
--- a/data/tmpfiles.d/image-builder.conf
+++ b/data/tmpfiles.d/image-builder.conf
@@ -1,0 +1,2 @@
+# Type  Path                        Mode UID  GID  Age
+d       /var/cache/image-builder    0755 root root 14d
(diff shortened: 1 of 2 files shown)


mvo5:
> systemd-rpm-macros

Thank you! My rpm-foo is weak :) I updated the spec file, *hope* I got it right with the %ghost dir as both ibcli and tmpfiles.d will create the dir as needed but if I should create it in the rpm that is fine, happy to do that too.

lzap:
One RPM build failed, but I am unable to download logs I see a timeout error:

https://dashboard.packit.dev/jobs/copr/3101972

PR #431, opened 2026-01-14
OSBuildOptions: Support --in-vm option

alexlarsson:


supakeen:
> Thanks for making the change.
> 
> However, this made me aware of the fact that we have a duplicate implementation of executing `osbuild` in this repository. Merging this PR would just add to the technical dept.
> 
> Would you be willing to look at actually replacing this with the implementation in https://github.com/osbuild/images/blob/main/pkg/osbuild/osbuild-exec.go, which should have all the knobs that you need and hopefully also that ibcli needs?
> 
> If you don't have the capacity to do this, because this is arguably much more work, I'd be OK with merging this PR and addressing this duplication later.

I feel this is an 'us' problem so if @alexlarsson feels the same I'm good with merging this and then we fix the duplication :)

PR #442, opened 2026-01-27
Containerfile*: remove the GOPROXY override

ondrejbudai:
Fedora 43 reverted the patch to use GOPROXY=direct, and it now follows the standard https://proxy.golang.org,direct, so we can remove it from our Containerfiles.

See https://fedoraproject.org/wiki/Changes/golang1.25


bcl (approved):
Looks good. The version of go in `registry.fedoraproject.org/fedora:43` is v1.25.6

croissanne (approved):

PR #444, opened 2026-01-28
verbose logging fixes

ondrejbudai:
2 commits meant for fixing, and improving the `--verbose` mode. Turns out it didn't work at all before.

**main: fix --verbose logging**

rootCmd.PersistentFlags().GetBool("verbose") always returned false prior
this commit because flags are not parse yet. This happens only when
rootCmd.Execute() is called.

Fix this by using one of the cobra hooks that runs after flags are
parsed. Note that the hook is needed because we want to enable verbose
logging for all subcommands. Otherwise, we would have to enable it in
every single subcommand.

I also added a small log message that actually informs users that
verbose logging is enabled.

---

**main: enable logrus in --verbose mode**

osbuild/images still use logrus for some its logging, so let's enable it
as well until we can fully migrate over.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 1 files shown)


lzap (changes requested):
The `logrus` dependency was pulled into `images` accidentally. I already cleaned it from `logrus` and created `pkg/olog` proxy that is supposed to do the job. I suggest that we do not merge this and I fix `images` instead.

lzap:
For the record: Logrus is gone, banned and we also enabled `olog` packages both in cli and images.

PR #447, opened 2026-02-02
setup.ValidateHasContainerTags: support use in rootless containers

alexlarsson:
When running in a rootless container (made possible via https://github.com/osbuild/bootc-image-builder/pull/1189) then generally the per-user container store is passed to bc-i-b, both because that makes sense, but also because the container doesn't have permissions to read the host store. This is typically done by passing `-v ~/.local/share/containers/storage:/var/lib/containers/storage` when running bc-i-b in podman.

Unfortunately when this happens, podman complains that the store is in the wrong place:

```
Error: database static dir "/home/alex/.local/share/containers/storage/libpod" does not match our static dir "/var/lib/containers/storage/libpod": database configuration mismatch
```

To fix this, we pass in podman arguments `--root` and `--imagestore` to set up an environment where the graphroot is some empty directory that has the "correct" location and we only load the image layers from /var/lib/containers.

To ensure this doesn't weirdly affect the regular case, this is all protected by podmanutil.IsRootless().

### DIFF
--- a/pkg/setup/setup.go
+++ b/pkg/setup/setup.go
@@ -152,7 +152,25 @@ func validateCanRunTargetArch(targetArch string) error {
 }
 
 func ValidateHasContainerTags(imgref string) error {
-	output, err := exec.Command("podman", "image", "inspect", imgref, "--format", "{{.Labels}}").Output()
+	extraOpts := []string{}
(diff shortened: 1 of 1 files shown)


bcl:
Do we have any tests that cover this?

PR #485, opened 2026-03-26
Support --in-vm in image-builder (HMS-10299)

avitova:
This adds the possibility of running the container in a VM. I tried to test this. The image is built. So I assume this is working? If anyone has any suggestions on how to test this, they are welcome.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -18,7 +18,7 @@ RUN mkdir -p /etc/containers/networks
 # Fast-track osbuild so we don't depend on the "slow" Fedora release process to implement new features in bib
 RUN dnf install -y dnf-plugins-core \
     && dnf copr enable -y @osbuild/osbuild \
-    && dnf install -y libxcrypt-compat wget osbuild osbuild-ostree osbuild-depsolve-dnf osbuild-lvm2 openssl subscription-manager \
+    && dnf install -y libxcrypt-compat wget osbuild osbuild-ostree osbuild-depsolve-dnf osbuild-lvm2 osbuild-virt-deps openssl subscription-manager \
     && dnf clean all
 
 COPY --from=builder /build/bin/image-builder /usr/bin/
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -415,13 +415,21 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	runInVm, err := cmd.Flags().GetBool("in-vm")
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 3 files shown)


Containerfile
```diff
@@ -19,6 +19,7 @@ RUN mkdir -p /etc/containers/networks
 RUN dnf install -y dnf-plugins-core \
     && dnf copr enable -y @osbuild/osbuild \
     && dnf install -y libxcrypt-compat wget osbuild osbuild-ostree osbuild-depsolve-dnf osbuild-lvm2 openssl subscription-manager \
+        qemu-system-x86 \
```
supakeen: In `bootc-image-builder` we install `qemu-kvm-core` instead.

bcl: That's because it also needs `virtiofsd` but osbuild didn't add any Requirements for this feature. The right way to do this is to add them to osbuild, or add a metadata-only subpackage like 'osbuild-in-vm' that Requires them. Otherwise we're going to be maintaining package requirements in more than one place :)

avitova:
Changing this to draft, since the other PR would have to go first. :)

avitova:
This is it. But I believe we might be installing maybe a little too much with `qemu` and `virtiofsd`. I think I kinda do not like that, as the usecase of the --in-vm is currently small, and we only support it in the container. Any thoughts? @supakeen

cmd/image-builder/main.go
```diff
@@ -415,15 +415,23 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
 	if setup.IsContainer() {
-		if err := setup.EnsureEnvironment(cacheDir, false); err != nil {
+		if err := setup.EnsureEnvironment(cacheDir, runInVm); err != nil {
 			return fmt.Errorf("entrypoint setup failed: %w", err)
 		}
 	}
 
+	if runInVm && !setup.IsContainer() {
+		return fmt.Errorf("running in VM outside container is not supported yet")
+	}
```
lzap, line 433: Forward looking statement warning :-) I mean, it is good to be positive I guess so why not...

lzap (approved):
Looks clean, haven't tried it yet.

Are we planning on updating the README soon? This is an exciting feature we should advertise and celebrate! :-D

supakeen:
> Looks clean, haven't tried it yet.
> 
> Are we planning on updating the README soon? This is an exciting feature we should advertise and celebrate! :-D

Since it's only applicable in a very small subset of use-cases that would need to be documented (e.g. you need to match OS between host and target *and* it only works inside containers). It's documented as experimental in `bootc-image-builder`; we could do the same snippet here.

supakeen (approved):

avitova:
Thank you!:)

PR #503, opened 2026-04-24
docs: update satellite info

lzap:
@jlsherrill noticed that we only recommend HTTP repos but this should work also on protected repos, this was tested. Updating the paragraph.


thozza (approved):

brlane-rht (approved):
Thanks!

PR #504, opened 2026-04-28
docs: generate manpages

supakeen:
Create manpages by using Cobra's built-in ability to do so. Add a hidden command that we call in our RPM specfile to generate the pages and install them in the appropriate location.

Also adds a new Makefile option to generate the manpages.

---

Looks like:

image
image

Etc.

### DIFF
--- a/cmd/image-builder/manpages.go
+++ b/cmd/image-builder/manpages.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/cobra/doc"
+)
+
+func addDocCmd(rootCmd *cobra.Command) {
+	docCmd := &cobra.Command{
+		Use:    "doc <output-dir>",
+		Short:  "Generate man pages for this command",
+		Hidden: true,
+		Args:   cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			header := &doc.GenManHeader{
+				Section: "1",
+			}
+			return doc.GenManTree(rootCmd, header, args[0])
+		},
+	}
+	rootCmd.AddCommand(docCmd)
+}
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ __pycache__
 *.tar.bz2
 recording-*.cast.json
 /image-builder
+man
--- a/Makefile
+++ b/Makefile
@@ -109,9 +109,14 @@ build: $(BUILDDIR)/bin/  ## build the binary from source
 	    GOARCH="$$arch" go build -ldflags="-s -w" -o ./bin/bib-canary-"$$arch" ./cmd/cross-arch/; \
 	done
 
+.PHONY: man
+man: build $(BUILDDIR)/man/man1/  ## Generate man pages
+	$(BUILDDIR)/bin/image-builder doc $(BUILDDIR)/man/man1/
+
 .PHONY: clean
 clean:  ## Remove all built binaries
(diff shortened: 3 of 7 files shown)


supakeen:
Mmm; seems to fail with on CentOS/RHEL, but not Fedora.

```
+ '%{gobuilddir}/bin/image-builder' gen-man-pages man/man1/
/var/tmp/rpm-tmp.r2wsy8: line 64: fg: no job control
```

Wonder what the actual error is.

supakeen:
> Mmm; seems to fail with on CentOS/RHEL, but not Fedora.
> 
> ```
> + '%{gobuilddir}/bin/image-builder' gen-man-pages man/man1/
> /var/tmp/rpm-tmp.r2wsy8: line 64: fg: no job control
> ```
> 
> Wonder what the actual error is.

It means that the macro isn't expanded at all; maybe we're lacking a buildrequires for Go macros or the version in EL doesn't contain this macro.

lzap (dismissed):
Looks good, not sure what you mean by the macro not being expanded, it is used in the file already why it would not error out there?

supakeen:
> Looks good, not sure what you mean by the macro not being expanded, it is used in the file already why it would not error out there?

Mmm; you're right. Confusing. Maybe it's only available under `%install` on RHEL I don't know if macros can do that.

cmd/image-builder/manpages.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/cobra/doc"
+)
+
+func addManPageCmd(rootCmd *cobra.Command) {
+	manCmd := &cobra.Command{
+		Use:    "gen-man-pages <output-dir>",
```
brlane-rht: For whatever it's worth I called this command `doc` in composer-cli. But I don't have strong feelings either way, other than consistency would be a good idea.

brlane-rht:
Yeah, it's not getting expanded on RHEL :) From a different PR where it works on accident:
```
+ install -m 0755 -vp '%{gobuilddir}/bin/image-builder' /builddir/build/BUILDROOT/image-builder-62-1.20260428083453660281.pr462.6.gc430891.el10.x86_64/usr/bin/
'%{gobuilddir}/bin/image-builder' -> '/builddir/build/BUILDROOT/image-builder-62-1.20260428083453660281.pr462.6.gc430891.el10.x86_64/usr/bin/image-builder'
```

You could just write it to `.` in gobuild and install and run from there instead.

supakeen:
> Yeah, it's not getting expanded on RHEL :) From a different PR where it works on accident:
> 
> ```
> 
> + install -m 0755 -vp '%{gobuilddir}/bin/image-builder' /builddir/build/BUILDROOT/image-builder-62-1.20260428083453660281.pr462.6.gc430891.el10.x86_64/usr/bin/
> 
> '%{gobuilddir}/bin/image-builder' -> '/builddir/build/BUILDROOT/image-builder-62-1.20260428083453660281.pr462.6.gc430891.el10.x86_64/usr/bin/image-builder'
> 
> ```
> 
> 
> 
> You could just write it to `.` in gobuild and install and run from there instead.

Ah because it's quoted in the other case. Ok I'll change both of them. Thanks 🙂

cmd/image-builder/manpages.go
supakeen: I'll call it doc.

lzap (approved):

brlane-rht (approved):
Looks good!

PR #505, opened 2026-04-28
main: `--version` -> `version` subcommand

supakeen:
Deprecate the `--version` subcommand because it mixes badly with wanting to pick an output format for it (it'd be easier to read JSON from our `koji-image-builder` plugin in the future).

Instead deprecate the old argument and introduce a new `version` subcommand with pickable output formats. Docs and tests included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,25 @@ $ image-builder list --format=json
 ]
 ```
 
+### Version information
+
+To show version information about the installed `image-builder`:
+```console
+$ image-builder version
+image-builder:
+  version: 0.5
+  commit: abc123
+  dependencies:
+    images: v0.100.0
+    osbuild: "105"
+```
+
+The output format can be changed with `--format`, supported formats are `yaml` (default) and `json`:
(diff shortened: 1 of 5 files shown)


lzap (approved):
Ok but do we really need YAML and JSON? Looks easy now, a commitment for the future.

supakeen:
> Ok but do we really need YAML and JSON? Looks easy now, a commitment for the future.

I'd have prefered plaintext *or* JSON from the start as that's something we can read in the Python ecosystem without external dependencies in (for example) `koji-image-builder` 🙁

brlane-rht (approved):
Since we're switching to a new command anyway why not make the default output more human friendly? I guess at least it's not json though.

supakeen:
> Since we're switching to a new command anyway why not make the default output more human friendly? I guess at least it's not json though.

I think at the time (because we wanted to show multiple versions) we decided that YAML was plain text readable for users and I still agree with that 🙂

PR #510, opened 2026-05-07
cmd: drop "bootc is experimental"

supakeen:
This is a left over that we should no longer be showing. Both PRs have been merged and it's especially confusing now that we're publically stating our timelines and asking users to migrate.

---

Mentioned by Niklas Reisser in https://github.com/osbuild/bootc-image-builder/issues/1221


lzap (approved):

brlane-rht (approved):
Ack!

PR #511, opened 2026-05-07
main: default to XDG cache directory for non-root users

### LINKED ISSUE #76: Handle permissions better

guillermodotn:
### Summary

- Default the `--cache` path to `$XDG_CACHE_HOME/image-builder/store` (or `~/.cache/image-builder/store`) when running as a non-root user, instead of the root-only `/var/cache/image-builder/store`
- Add an early writability check on the cache directory to fail fast with an actionable hint, instead of waiting for osbuild to fail after manifest generation

Fixes #76

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -38,13 +38,12 @@ var (
 	osStderr io.Writer = os.Stderr
 )
 
-// defaultCacheDir returns the default cache directory for osbuild
-// intermediate build artifacts. When running as root it uses the
-// system-wide /var/cache path. When running as a non-root user it
-// follows the XDG Base Directory specification and falls back to
-// ~/.cache.
-func defaultCacheDir() string {
-	if os.Getuid() == 0 {
+// cacheDirForUid returns the cache directory for the given uid.
+// When root (uid 0) it uses the system-wide /var/cache path.
+// When non-root it follows the XDG Base Directory specification
+// and falls back to ~/.cache.
+func cacheDirForUid(uid int) string {
+	if uid == 0 {
 		return "/var/cache/image-builder/store"
 	}
 	if cacheHome := os.Getenv("XDG_CACHE_HOME"); cacheHome != "" {
@@ -55,6 +54,11 @@ func defaultCacheDir() string {
 		return "/var/cache/image-builder/store"
 	}
 	return filepath.Join(home, ".cache", "image-builder", "store")
+}
+
+// defaultCacheDir returns the cache directory for the current user.
+func defaultCacheDir() string {
+	return cacheDirForUid(os.Getuid())
 }
 
 // basenameFor returns the basename for directory and filenames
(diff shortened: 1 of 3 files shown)


supakeen (dismissed):
Thank you; this is pretty much what I had in mind for this.

Note that we do ship a systemd tmpfiles dropin to clean up `/var/cache/image-builder` but we don't do this for the new locations. You don't need to address this but I did want to mention it in this PR.

In the future we should grow a `image-builder cache ...` subcommand to manage/prune the cache.

brlane-rht (changes requested):
I'd like to see tests that always run since right now they are going to miss coverage depending on how they are invoked. The simple solution is to pass in os.Getuid to DefaultCacheDir, but that's kinda ugly.

cmd/image-builder/main.go
```diff
@@ -440,7 +460,11 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	// XXX: check env here, i.e. if user is root and osbuild is installed
+	// Fail early if the cache directory is not writable, instead of
+	// waiting for osbuild to fail after slow manifest generation.
+	if err := os.MkdirAll(cacheDir, 0755); err != nil {
+		return fmt.Errorf("cannot create cache directory %q: %w\nHint: use --cache to specify a writable path", cacheDir, err)
+	}
```
lzap, line 465: This is nice.

lzap (approved):

brlane-rht (approved):
Thanks!

PR #512, opened 2026-05-08
main: hide `--with-rpmlist`

supakeen:
This option is really only used by Koji and shouldn't clutter up the overviews for normal users. Let's hide it.


brlane-rht (approved):
Looks good. Is there a way to list all the hidden options? I couldn't find one in the Cobra docs.

thozza (approved):

PR #513, opened 2026-05-08
main: hide `--use-librepo`

supakeen:
This is an advanced feature that should only be used if the default behavior actually causes issues. `osbuild` versions that support this have been deployed everywhere as well.


brlane-rht (approved):
Ack

thozza (approved):

PR #516, opened 2026-05-13
cmd: introduce profiling options

lzap:
Add persistent flags --memprofile, --memprofile-goroutine, and --memprofile-rate. Heap and goroutine profiles are written in Go pprof format when the process exits, including on command failure, so runs can be analyzed with the toolchain from go.mod (e.g. go tool pprof).

When profiling is off, allocation sampling is not enabled for the image-builder CLI: MemProfileRate is reset at the start of run() and stays zero unless a command sets --memprofile, in which case MemProfileRate follows --memprofile-rate (-1 means Go's default 524288).

```
image-builder -v manifest ... --memprofile /tmp/heap.pprof \
    --memprofile-goroutine /tmp/goroutine.pprof
```

```
go tool pprof -http=:8080 /tmp/heap.pprof
```

### DIFF
--- a/cmd/image-builder/memprofile.go
+++ b/cmd/image-builder/memprofile.go
@@ -1,0 +1,89 @@
+//go:build profiling
+
+// See @cmd/image-builder/memprofile_stub.go for API documentation shared with the non-profiling build.
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"runtime"
+	"runtime/pprof"
+
+	"github.com/spf13/cobra"
+)
+
+// Go default allocation sampling rate (see runtime.MemProfileRate).
+const defaultMemProfileRate = 512 * 1024
+
+var (
+	memProfilePath          string
+	memProfileGoroutinePath string
+	memProfileRateOpt       int = -1
+)
+
+func memProfileResetForProcessStart() {
+	runtime.MemProfileRate = 0
+}
+
+func registerMemProfileFlags(root *cobra.Command) {
+	f := root.PersistentFlags()
+	f.StringVar(&memProfilePath, "memprofile", "", "write a heap memory profile in pprof format when the program exits (use with go tool pprof)")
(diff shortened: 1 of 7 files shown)


supakeen (commented):
I'd like to have these things but like @croissanne says they should be behind a build flag/tag so we don't ship this in production executables.

cmd/image-builder/memprofile.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func registerMemProfileFlags(root *cobra.Command) {
+	f := root.PersistentFlags()
+	f.StringVar(&memProfilePath, "memprofile", "", "write a heap memory profile in pprof format when the program exits (use with go tool pprof)")
+	f.StringVar(&memProfileGoroutinePath, "memprofile-goroutine", "", "write a goroutine profile in pprof format on exit (optional; use with go tool pprof)")
+	f.IntVar(&memProfileRateOpt, "memprofile-rate", -1, "when --memprofile is set, sets runtime.MemProfileRate for allocation sampling (-1 uses Go's default 524288; 0 samples only live heap at exit with minimal runtime overhead)")
+}
+
+func memProfilePersistentPreRun(_ *cobra.Command, _ []string) {
```
brlane-rht, line 37: This is a bit hard to read. It looks like the rate is always 0 unless `memProfilePath` is used, so `memProfileGoroutinePath` doesn't matter. I'd unconditionally set it to 0 first, then override that with the `if memProfilePath != "" {` block.

brlane-rht (dismissed):
Looks good other than my one small comment.

cmd/image-builder/memprofile.go
lzap: Yeah thanks, rebased.

brlane-rht (approved):
Looks good.

supakeen (approved):

PR #525, opened 2026-05-25
test: switch to Fedora 44

supakeen:
Remove all mentions of Fedora 42 in our tests and move to Fedora 44 where applicable.

I noticed in some PRs that tests based on Fedora 42 started failing, they pass on 44 and 42 is EOL now-ish.


thozza (approved):

brlane-rht (approved):
Looks good!

PR #529, opened 2026-05-26
main: `bootc inspect`

supakeen:
Introduce a new `bootc` subcommand where we can grow various helpers for working with bootable containers.

The first one of these is the `inspect` subcommand, this allows a user to dump the struct we get from `images` to determine what `image-builder` thinks about the container.

Usage as follows:

```
€ sudo ./image-builder bootc inspect --format json --ref localhost/fedora-kinoite-iso:44 | jq .OSInfo.KernelInfo
{
  "Version": "7.0.10-200.fc44.x86_64",
  "HasAbootImg": false
}
```

---

As said in the documentation this is an internal structure but it's still useful. Perhaps in the future we can standardize it.

The intent is also to have an umbrella to add additional `bootc` subcommands to with us /eventually/ migrating away from `--bootc-*` arguments on `build` to having them on `bootc convert` though that's a very big IF/MAYBE/PERHAPS.

### DIFF
--- a/cmd/image-builder/bootc_test.go
+++ b/cmd/image-builder/bootc_test.go
@@ -1,0 +1,141 @@
+package main_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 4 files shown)


brlane-rht (approved):
Looks good!

lzap (approved):

PR #530, opened 2026-05-26
doc: mention additional requirements

supakeen:
More files are required for a `bootc-generic-iso` build to succeed. In the past most containers had `grub2` in them in its entirety. With the advent of sealed containers there are now containers out there that only have `systemd-boot` by default.

---

For osbuild/image-builder#2423.


brlane-rht (approved):
Looks ok for now. I'll have an example Containerfile to go along with the change to booting using ostree instead of the deployed copy.

lzap (approved):

PR #534, opened 2026-05-29
main: `--force-defs-dir`

supakeen:
Previously it was possible to pass a custom definition directory through the use of `IMAGE_BUILDER_EXPERIMENTAL` options. In `images` 0.269.0 I landed a PR [1] that allows us setting the definition path in code.

This commit implements that, plus test cases, to pass custom definitions.

We keep the argument hidden as we still consider it to be experimental until we flesh out certain things but this *will* allow us to source our default image definitions from disk instead of the embedded filesystem in the near future (a follow up) so consider this the lead-up to that.

The argument follows the same naming as others, `force` means to entirely wipe and override the built-ins. In the future there will likely be an `extra` equivalent to extend, or override pre-existing definitions in the same way we do for repositories.

We might enable this for build systems in their respective plugins at some point as well.

[1]: https://github.com/osbuild/images/pull/2378


lzap (approved):

brlane-rht (approved):
Looks good!

PR #537, opened 2026-06-01
main: `system` subcommand

supakeen:
Let's add a `system` subcommand that (initially) shows information about the cache location, its max-size and its current size.

This will be a good umbrella subcommand to hang other commands under such as `prune` to clear cache in the future. 

```
€ sudo ./image-builder system
system:
  cache:
    path: /var/cache/image-builder/store
    size: 388024255
    max-size: unlimited
```

### DIFF
--- a/cmd/image-builder/system.go
+++ b/cmd/image-builder/system.go
@@ -3,22 +3,92 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
 
 	"go.yaml.in/yaml/v3"
 )
 
+type cacheMaxSize struct {
+	set   bool
+	value int64
+}
+
+func (c cacheMaxSize) MarshalJSON() ([]byte, error) {
+	if !c.set {
+		return json.Marshal("unknown")
+	}
+	if c.value == 0 {
+		return json.Marshal("unlimited")
+	}
+	return json.Marshal(c.value)
+}
+
+func (c cacheMaxSize) MarshalYAML() (interface{}, error) {
+	if !c.set {
+		return "unknown", nil
+	}
+	if c.value == 0 {
+		return "unlimited", nil
+	}
+	return c.value, nil
+}
+
 type systemStatus struct {
 	System struct {
 		Cache struct {
-			Path string `yaml:"path" json:"path"`
+			Path    string       `yaml:"path" json:"path"`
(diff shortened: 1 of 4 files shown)


cmd/image-builder/system.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (c cacheMaxSize) MarshalJSON() ([]byte, error) {
+	if !c.set {
+		return json.Marshal("unknown")
+	}
+	if c.value == 0 {
+		return json.Marshal("unlimited")
+	}
+	return json.Marshal(c.value)
```
avitova, line 27: This makes negative values a valid number, does it not?

cmd/image-builder/system.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (c cacheMaxSize) MarshalYAML() (interface{}, error) {
+	if !c.set {
+		return "unknown", nil
+	}
+	if c.value == 0 {
+		return "unlimited", nil
+	}
+	return c.value, nil
```
avitova, line 37: Same question here for negative numbers. Do we want to handle that? Or should we make cacheMaxSize uint?

avitova (commented):
Just a small question. 🤏

cmd/image-builder/system.go
```diff
@@ -3,22 +3,92 @@ package main
...
+			info, err := d.Info()
+			if err != nil {
+				return err
+			}
+			total += info.Size()
+		}
+		return nil
+	})
+	return total, err
+}
```
achilleas-k, line 69: osbuild maintains a file in the root of the cache with the current size, so we could read that.

achilleas-k: Though it seems it's updated by calculating writes and deletes, not measuring.  I checked mine right now and it's off from the real size on disk by quite a bit.  Perhaps a bug for osbuild.

cmd/image-builder/system.go
supakeen: Yea, a negative value is valid in this case (though I don't know what the `osbuild` side does).

cmd/image-builder/system.go
supakeen: Both are a bit buggy I think, on the `osbuild` side and the code in this PR. Especially if mountpoints are set up.

achilleas-k (approved):

brlane-rht (approved):
I would argue that `prettySystemStatus` outputting yaml doesn't live up to its name, but it is better than json :)

